PR #6, opened 2021-10-07
Fix SeaExplorer L0 timeseries name to avoid broken website links

hvdosser:
Removed _L0 from the end of the SeaExplorer L0 timeseries name for the netcdf files. This change makes the filename consistent with the L0 timeseries filename for the Slocums, and avoids creating broken website links, since the html template is looking for a filename without L0 appended to the end.


jklymak:
@hvdosser you will probably want to rebase on upstream/main to get CI to work here.

Issue #7, opened 2021-10-21
xarray error adding time attribute in seaexplorer


callumrollo:
When running the seaexplorer example, xarray throws an error in seaexplorer.py line 361:

```python
INFO:pyglider.seaexplorer:interpolating latitude
Traceback (most recent call last):
  File "/home/callum/Documents/data-flow/raw-to-nc/pyglider/example-seaexplorer/process_deploymentRealTime.py", line 35, in <module>
    outname = seaexplorer.raw_to_L0timeseries(rawncdir, l0tsdir, deploymentyaml, kind='sub')
  File "/home/callum/Documents/data-flow/raw-to-nc/pyglider/pyglider/seaexplorer.py", line 361, in raw_to_L0timeseries
    ds[name] = (('time'), val, attrs)
  File "/home/callum/anaconda3/envs/pyglider/lib/python3.9/site-packages/xarray/core/dataset.py", line 1563, in __setitem__
    self.update({key: value})
  File "/home/callum/anaconda3/envs/pyglider/lib/python3.9/site-packages/xarray/core/dataset.py", line 4208, in update
    merge_result = dataset_update_method(self, other)
  File "/home/callum/anaconda3/envs/pyglider/lib/python3.9/site-packages/xarray/core/merge.py", line 984, in dataset_update_method
    return merge_core(
  File "/home/callum/anaconda3/envs/pyglider/lib/python3.9/site-packages/xarray/core/merge.py", line 632, in merge_core
    collected = collect_variables_and_indexes(aligned)
  File "/home/callum/anaconda3/envs/pyglider/lib/python3.9/site-packages/xarray/core/merge.py", line 294, in collect_variables_and_indexes
    variable = as_variable(variable, name=name)
  File "/home/callum/anaconda3/envs/pyglider/lib/python3.9/site-packages/xarray/core/variable.py", line 121, in as_variable
    raise TypeError(
TypeError: Using a DataArray object to construct a variable is ambiguous, please extract the data using the .data property.

Process finished with exit code 1
```

I think this worked in earlier versions of xarray, but more recently the package is more strict on ambiguity. I used a conda environment created from the environment.yml included in the repo with python=3.9.7, xarray=0.19.0:

```shell
# packages in environment at /home/callum/anaconda3/envs/pyglider:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
[...]

PR #8, opened 2021-10-21
fix xarray time attribute error

callumrollo:
Fixes issue #7


jklymak:
The tests haven't been run for quite a while, so the new version of xarray being loaded is not compatible with the old code.  It'd be great if you could patch the slocum code as well.

callumrollo:
Nice! I'm happy for this to be merged whenever.

If I may be so bold, it could be worth adding test suites on other Python versions, at least 3.7 and 3.9. Testing on 3.10 is still a bit tricky atm but could be added later

jklymak:
Thanks @callumrollo !

Issue #22, opened 2021-10-26
Use of glider_devices


callumrollo:
Are the fields in glider_devices from the deployment yaml used anywhere? I couldn't find them being accessed by any of the scripts in gliderpy. These fields include sensor names, serial numbers and calibration information. Could be useful to include in netCDF metadata, but I don't know how that would be done. Could just append them to metadata?

soerenthomsen:
@callumrollo just reading through this now. Feel free to also join the debate about OG1.0 in the OceanGliders Data Format repo

PR #23, opened 2021-10-26
adding sensor data to netcdf globals

callumrollo:
To resolve Issue #22 

This is a very simplistic way to add the sensor data to netCDF metadata. This adds eash sensor as an attribute to the netcdf, with the dictionary of metadata included in a string.

To access it, the end user would do like:

`ctd_metadata = yaml.safe_load(ds.attrs['ctd'])`


jklymak (approved):
This seems reasonable to me.  Needs a rebase....

Issue #26, opened 2021-10-27
yaml to control sensor processing


callumrollo:
As discussed in PR #14, it would be good to control coarsening of data and perhaps other things? This could be done with optional additions to the deployment yaml. For discussion

callumrollo:
Resolved in PR #30

PR #30, opened 2021-10-29
specify timebase and coarsening in yaml

callumrollo:
Addressing Issue #26 

This allows the user to specify a sensor use as a timebase. If a sensor is not specified, it searches for gpctd or legato.

It also allows the option to coarsen data and set the coarsening on a per-variable basis. As is the case with present coarsening of oxygen, this is essentially a smoothing function, as all variables are on the same timebase.

This coarsening and timebase change is performed in `raw_to_L0timeseries`, so does not affect the creation of raw netcdf files. Positive: no data lost in conversion from csv to initial netcdfs. Negative: initial netcdfs will be large if sampling frequency is high


callumrollo:
Added capability to use "keeper variables" in the yaml. This should enable processing of full raw datasets without sacrificing sensor data or creating impractically large netcdfs.

To use this capability, the user should specify one of the variables that is always sampled, like `NAV_LONGITUDE` as the timebase. This will create a dataset with every timestep in raw.pld. This dataset is then reduced to the timestamps that have a non-blank entry for the sensors specified in deployment yaml  `netcdf_variables[ keep_variables]`.

The result is a dataset consisting of timestamps which have data for at least on of the instruments of interest. This contrasts the current system, which has only timestamps where the CTD registered a sample.

This has no effect on the processing of sub datasets (such as the example) as these include lines of data from every sensor

callumrollo:
I've made a demo of the effect this combined-dataset approach has:

https://github.com/callumrollo/combi-dataset-demo

callumrollo:
In a nutshell, this PR allows a user to use any number of sensors as a timebase. Only timepoints where one or more of these sensors recorded data are retained. This results in a final netCDF around 1/10th the size of the full dataset, with no loss of science data.

jklymak (approved):
This looks great!  Only issue is I wonder if we want to encourage explicitly specifying the timebase. maybe with a log.warning if its not there?  Or maybe that is adding too much to the users' overhead.

callumrollo:
I think encouraging explicit specification of a timebase is good. Adding it to the yml of the examples should achieve that.

PR #31, opened 2021-11-01
Add unit tests seaexplorer.py

callumrollo:
This PR will add at least one test for every function in seaexplorer.py


callumrollo:
All six functions that appear to be used now have tests. I can't find anywhere that:

- `_sort`
- `_interp_pld_to_pld`
- `raw_to_L1timeseries` (alias)

are used. recommend removal.

I added `_parse_sensor_config` by mistake in an earlier PR. I will remove it

Issue #15, opened 2021-10-25
Salinity calculation factor 10 error


callumrollo:
There's something weird going on with salinity. The factor of 10 multiplication of conductivity in `utils.py` line 200 wrecks our data, resulting in salinities up around 100 g/kg.

Is this factor of 10 necessary for processing GPCTD data? If so, it should be applied in place to conductivity. At present, the SX example produces final netcdfs where salinity is ~ 10x conductivity. I think this is most likely an error.

If this function is to be refactored, I would recommend moving away from [seawater](https://github.com/pyoceans/python-seawater/). This package has not been updated since 2015. The maintainer (Filipe) recommends moving to TEOS-10 as implemented in [gsw](https://github.com/TEOS-10/GSW-Python)

callumrollo:
PR #18 resolved the issue with the legato, which does not need to the 10x factor. However, the present solution only applies the 10x correction when calculation salinity for the GPCTD, conductivity is not itself corrected. This results in conductivity values around 3.5-4 in the example-seaexplorer final netcdf. I think these should be 35-40

callumrollo:
Resolved in #34

PR #34, opened 2021-11-05
apply 10* correction to gpctd conductivity

callumrollo:
This resolves Issue #15 

This fix changes all the conductivity values in the seaexplorer example, so will break the integration tests

### DIFF
--- a/pyglider/utils.py
+++ b/pyglider/utils.py
@@ -198,10 +198,11 @@ def get_profiles_new(ds, min_dp = 10.0, inversion=3., filt_time=100,
 
 def get_derived_eos_raw(ds):
     # GPCTD requires a scale factor of 10 for conductivity. Legato does not
-    if 'LEGATO' in ds.conductivity.source:
-        r = ds.conductivity / seawater.constants.c3515
-    else:
+    if 'GPCTD' in ds.conductivity.source:
         r = 10 * ds.conductivity / seawater.constants.c3515
+        ds['conductivity'] = ds['conductivity'] * 10
+    else:
+        r = ds.conductivity / seawater.constants.c3515
     ds['salinity'] = (('time'),
                       seawater.eos80.salt(r, ds.temperature, ds.pressure))
     attrs = collections.OrderedDict([('long_name', 'water salinity'),
--- a/pyglider/utils.py
+++ b/pyglider/utils.py
@@ -203,7 +203,8 @@ def get_derived_eos_raw(ds):
     elif 'mS cm' in ds.conductivity.units:
         r = ds.conductivity / seawater.constants.c3515
     else:
-        _log.error("Could not parse conductivity units in yaml. Expected 'S m-1' or 'mS cm-1'")
+        raise ValueError("Could not parse conductivity units in yaml. Expected 'S m-1' or 'mS cm-1'.\n"
+                         "Check yaml entry netcdf_variables: conductivity: units")
     ds['salinity'] = (('time'),
                       seawater.eos80.salt(r, ds.temperature, ds.pressure))
     attrs = collections.OrderedDict([('long_name', 'water salinity'),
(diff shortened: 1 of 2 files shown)


pyglider/utils.py
```diff
@@ -197,11 +197,10 @@ def get_profiles_new(ds, min_dp = 10.0, inversion=3., filt_time=100,
 
 
 def get_derived_eos_raw(ds):
-    # GPCTD requires a scale factor of 10 for conductivity. Legato does not
-    if 'LEGATO' in ds.conductivity.source:
-        r = ds.conductivity / seawater.constants.c3515
-    else:
-        r = 10 * ds.conductivity / seawater.constants.c3515
+    # GPCTD and slocum ctd require a scale factor of 10 for conductivity. Legato does not
+    if 'GPCTD' in ds.conductivity.source or 'sci_water_cond' in ds.conductivity.source:
```
jklymak, line 201: I'd still rather this was somehow in the yaml.  i.e. `conductivity.units` and then it is S/m for seabird, and mS/cm for RBR and we convert that way?

callumrollo: That would be a better way of implementing it. I've changed the logic now so it checks for units rather than sensor name. Should be more robust to new sensors

callumrollo:
The check now looks for "S m" or "mS cm" in the units. This won't catch alternative ways of wrting this like "S/m". It currently raises a log error if it can't parse the units. Hopefully that's enough for the end user to check and edit their yaml

pyglider/utils.py
```diff
@@ -197,11 +197,13 @@ def get_profiles_new(ds, min_dp = 10.0, inversion=3., filt_time=100,
...
-    # GPCTD requires a scale factor of 10 for conductivity. Legato does not
-    if 'LEGATO' in ds.conductivity.source:
+    # GPCTD and slocum ctd require a scale factor of 10 for conductivity. Legato does not
+    if 'S m' in ds.conductivity.units:
+        r = 10 * ds.conductivity / seawater.constants.c3515
+    elif 'mS cm' in ds.conductivity.units:
         r = ds.conductivity / seawater.constants.c3515
     else:
-        r = 10 * ds.conductivity / seawater.constants.c3515
+        _log.error("Could not parse conductivity units in yaml. Expected 'S m-1' or 'mS cm-1'")
```
jklymak, line 206: should this actually raise?   I don't think errors in the log raise, and if they don't then r won't be defined, and it will fail below.

callumrollo: Testing it looks like this does not raise. Would it be better to raise a ValueError if it can't match either of the clauses?

jklymak: I think so with a robust error message that tells folks to edit the yml?

callumrollo: Sounds good, implemented. it now raises a ValueError if it can't parse the conductivity units

pyglider/utils.py
```diff
@@ -197,11 +197,14 @@ def get_profiles_new(ds, min_dp = 10.0, inversion=3., filt_time=100,
...
-    # GPCTD requires a scale factor of 10 for conductivity. Legato does not
-    if 'LEGATO' in ds.conductivity.source:
+    # GPCTD and slocum ctd require a scale factor of 10 for conductivity. Legato does not
+    if 'S m' in ds.conductivity.units:
+        r = 10 * ds.conductivity / seawater.constants.c3515
+    elif 'mS cm' in ds.conductivity.units:
         r = ds.conductivity / seawater.constants.c3515
     else:
-        r = 10 * ds.conductivity / seawater.constants.c3515
+        raise ValueError("Could not parse conductivity units in yaml. Expected 'S m-1' or 'mS cm-1'.\n"
```
jklymak, line 206: I'm not sure what conventions we want to have, but the gurus in Matplotlib tell me not to put our own carriage returns in Errors and warnings.  I'm not too fussed if we do this or not, but they claim it improves readability if someone has a narrow terminal or something.

jklymak (approved):

pyglider/utils.py
callumrollo: Good to follow best practices. CR removed

PR #35, opened 2021-11-05
improve test granularity: one test for each variable of dataset

callumrollo:
This makes use of pytest parameterise to test each variable in the dataset individually. This helps diagnosis of problems e.g. in PR #34 where one variable in the ds is changing to check if it affects others too

### DIFF
--- a/tests/test_pyglider.py
+++ b/tests/test_pyglider.py
@@ -1,27 +1,34 @@
 import xarray as xr
 from pathlib import Path
 import sys
+import pytest
 
 library_dir = Path(__file__).parent.parent.absolute()
 sys.path.append(str(library_dir))
 import pyglider.seaexplorer as seaexplorer
 import pyglider.slocum as slocum
 
+# Create an L0 timeseries from seaexplorer data and test that the resulting netcdf is identical to the test data
+rawdir = str(library_dir / 'example-seaexplorer/realtime_raw/') + '/'
+rawncdir = str(library_dir / 'example-seaexplorer/realtime_rawnc/') + '/'
+deploymentyaml = str(library_dir / 'example-seaexplorer/deploymentRealtime.yml')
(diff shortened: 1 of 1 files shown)


tests/test_pyglider.py
```diff
@@ -1,27 +1,34 @@
...
+rawncdir = str(library_dir / 'example-seaexplorer/realtime_rawnc/') + '/'
+deploymentyaml = str(library_dir / 'example-seaexplorer/deploymentRealtime.yml')
+l0tsdir = str(library_dir / 'example-seaexplorer/L0-timeseries-test/') + '/'
+seaexplorer.raw_to_rawnc(rawdir, rawncdir, deploymentyaml)
+seaexplorer.merge_rawnc(rawncdir, rawncdir, deploymentyaml, kind='sub')
+outname = seaexplorer.raw_to_L0timeseries(rawncdir, l0tsdir, deploymentyaml, kind='sub')
+output = xr.open_dataset(outname)
+# Open test data file
+test_data = xr.open_dataset(library_dir / 'tests/results/example-seaexplorer/L0-timeseries/dfo-eva035-20190718.nc')
+variables = list(test_data.variables)
```
jklymak, line 22: OK, this doesn't test if we drop a whole variable by accident, does it?  Maybe you should add a check that the variable list is the same for each file?  

Also pretty sure `rawdir` etc can remain as `Path` objects, rather than converting to strings?

callumrollo: I'll add a test to make sure we haven't dropped variables. At the moment, `rawdir` and others have to be passed as paths, or it errors out with stuff like `d = indir + f'*.{ftype}.{rawsub}.*'` in raw_to_rawnc. We could use refactoring the whole module to use paths rather than strings really.

jklymak (approved):

jklymak:
Feel free to merge when CI passes...

Issue #36, opened 2021-11-05
Slow processing of large datasets


callumrollo:
There appears to be a performance bottleneck in raw_to_rawnc. In the docstring it is noted that this is slow for large datasets. As this function converts small files individually, it looks ideal for multiprocessing.

callumrollo:
Update: I've recently been introduced to [polars](https://github.com/pola-rs/polars) which looks like it would be an ideal solution for this. It's parallel by default  and works effectively as a replacement for pandas. I'll work up a PR to use polars in place of pandas for the seaexplorer raw_to_rawnc step

jklymak:
We also introduced a method to subsample the data to remove all the redundant data the Alseamars put out by default.  Not sure if that helps fix the problem.  Not to say we shouldn't also consider polars.

PR #37, opened 2021-11-05
multiprocessing seaexplorer raw to rawnc

callumrollo:
This resolves Issue #36 

The multiprocessing defaults to using 80 % of cores available on the host machine. I'm not sure what best practice is on this, but it works very will with my 16 core workstation. Speeds up processing of large datasets considerably. The setting of core numbers is always rounded down, so it will only use 1 core on a 2 core machine.

Specifying cores=1 will result in processing identical to the original setup. Happy to make this the default if preferred.

I'll update the tests to cover this new functionality

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -13,7 +13,10 @@ import xarray as xr
 import yaml
 import pyglider.utils as utils
 import pandas as pd
-
+import random
+import math
+import multiprocessing as mp
+import time
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


callumrollo:
In the hour or so it took to write this patch and process a dataset with multiproc, the original processing run with a single core has still not completed :1st_place_medal: Definitely worth improving

jklymak:
This seems useful. OTOH do we know why it is so slow?

callumrollo:
The process to convert input files into netcdfs appears to be CPU bound. It doesn't take much memory, but Python's GIL that limits it to one process means only one core works at a time, unless you use something like multiprocess.

Given that we want to convert all the input csvs into netcdfs, I think this is the best way to speed it up. Processing time will decrease proportionally to the number of cores you can throw at it.

callumrollo:
Alternatively, it could be an IO bottleneck. If so, the solution of processing multiple files in parallel should still be the best optimisation route.

I haven't done much Python profiling before, but I can give it a go.

jklymak:
I _think_ its because the files are so big, if I recall.  Certainly I don't at all object to multiprocessing, but just wanted to be sure we weren't doing something dumb in reading or converting them that was taking too long.

callumrollo:
Largebig input files aren't a big problem at this step, as the parallelising allows one file to be read as another is processed etc. The big files do make `merge_rawnc` very slow though. As we're not doing the yaml-defined subsetting until `raw_to_L0timeseries` this step of combining many full size netcfs into one big one is time and memory consuming.

I'm not sure how best to address this. Do we drop some unwanted data in `raw_to¬rawnc`? Or when files are opened in `merge_rawnc`? Or do we just accept that processing raw data from a full mission will be a long memory intensive process that you hope to only do once?

jklymak:
Hmm ok maybe the merge raw part is the problem and we should consider _not_ making the raw merged file, but rather merge or concat the _timeseries_ files.  

For me getting the raw data to netcdf form is useful to check for problems. I'm not sure the raw merged netcdf is useful, though.  I would guess that merging the timeseries files with the redundant data removed would be faster, but I could be misremembering.

callumrollo:
Certainly making and keeping the per-dive netcdfs is valuable. If we don't need to create a full-sized raw merged netcdf, perhaps we could do some smarter processing. At the moment it's a for loop that appends to a dataset. Maybe something like [xarray open_mfdataset](http://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html) could help? In particular the `preprocess` kwarg looks like it could enable subsetting when the per-dive netcdfs are bing read in

callumrollo:
This commit moves the step of subsetting data by variable (PR #30) forward one step, so that it occurs as part of `merge_rawnc`. This substantially speeds up the merge step for datasets with more than ~ 50 dives. This means that the merged netcdf for the mission is also substantially smaller. The previous step is unaffected, so we're still creating per-dive netcdfs that have all the data from the raw files

callumrollo:
I'll add some tests for the new functionality and tidy it up a bit. If this is successful, we'll have multiprocessing on the two most time consuming steps

Issue #42, opened 2021-12-06
AROD oxygen data wrong


callumrollo:
From @bastienqueste:

> Note that oxygen from the AROD is wrong as it assumes a salinity of 0 PSU. To get proper oxygen, calculate saturation from the bad oxygen concentration and solubility calculated with good temperature and 0 salinity.
> From that saturation (which is a useful one to plot), we can get good concentration using a solubility calculated with good tmeperature and salinity.
> This salinity related error is probably about ~15-30% of the value so it's not worth ignoring

Does this fall within the remit of pyglider? It's not calibration of correction, just reporting accurate vales. It would involve a recalculation though.

jklymak:
I don't know much about O2.  However, I think pyglider should convert to as close-to-correct as possible...

jklymak:
(so long as the processing steps are able to be backed out from raw files)

bastienqueste:
Careful though - people can change the onboard settings on the sensor to specify their own salinity - which some people do.
You can't assume it's 0 onboard and there's no way to know.
What Alsemar need to do is start outputting engineering units.

richardsc:
@bastienqueste agree about how confusing this can be. It took us a long time of asking Alseamar and JFE before we even figured out the details that you mentioned above.

I believe that the assumption of 0 salinity for the AROD is the default, so perhaps that would be a good thing for pyglider to assume but with a caution for the user that they should check the configuration of their sensor. There is no way to know from the glider data what it is without connecting to the AROD separately and using the sensor software.

Note that this is also potentially an issue with the Aanderra optode normally mounted on slocum gliders -- there is a default salinity assumed in the output O2 concentration that should be corrected following the same procedure (convert to saturation using the assumed salinity and then recalculate concentration using the actual salinity/temperature). I don't know what the default for Aanderra is, but I recall it might be more sensible (like 34?).

bastienqueste:
Same issue for all optodes. Coda included.

Aanderaa defaults change based on location specified on order and when it was. We've had some with different settings when we used to order with seagliders.

I agree proceeding with a warning is a viable option.

In a couple iterations, most or all sensors will start producing their own .raw files like the legato has started doing. These should have all data columns.

callumrollo:
Resolved in #43

PR #43, opened 2021-12-08
Correction to arod oxygen

callumrollo:
This PR addresses Issue #42 

Correction to oxygen from @bastienqueste. This correction is applied if the user has a variable called correct_oxygen in the oxygen_concentration section of deployment yaml. It looks for the variable reference_salinity in the same section and uses this as the reference when correcting for oxygen. If no reference salinty is found, 0 is assumed. This generates a Warning


callumrollo:
This correction is placed further down than others (like nmea2deg) as it requires other variables like temperature to be standardised first.

This adds **gsw** as a requirement.

callumrollo:
image
Here's the old (blue) and corrected (orange) oxygen concentration from **example-seaexplorer-legato-flntu-arod-ad2cp**. A non-negligible change.

callumrollo:
@bastienqueste could you have a look at this and check the salinity correction has been implemented correctly?
 
 @richardsc does this use of warnings seem sensible? There's functionality to change the reference salinity in the yaml

richardsc:
@callumrollo I'm happy with both the warnings (haven't tried to run it myself to see what it looks like, but the messages as entered in the code look good) and the yaml parameters for specifying the reference.

pyglider/utils.py
```diff
@@ -361,6 +362,21 @@ def nmea2deg(nmea):
...
 
+
+def oxygen_concentration_correction(data, ncvar):
+    oxy_yaml = ncvar['oxygen_concentration']
+    if 'reference_salinity' not in oxy_yaml.keys():
+        _log.warning('No reference_salinity found in oxygen deployment yaml. Assuming reference salinity of 0 psu')
+        ref_sal = 0
+    else:
+        ref_sal = float(oxy_yaml['reference_salinity'])
+    sa = gsw.SA_from_SP(data['salinity'], data['pressure'], data['longitude'], data['latitude'])
```
jklymak, line 373: OK, so maybe I should have read the gsw PR more carefully.  What are longitude and latitude doing in this call?

richardsc: Ah good catch Jody. While the glider data should only store SP, of course the `gsw_O2sol()` function requires SA and CT (which in turn require lon/lat). See:

https://www.teos-10.org/pubs/gsw/html/gsw_O2sol.html

GSW is great but man sometimes it fixes something that wasn't really broken in the first place ...

jklymak: Oh, I see, so `data['salinity']` is indeed practical salinity, or at least some version that is not geographically corrected, but O2sol wants the correction.  Though I'm not sure we want to give it the corrected version?  What happens if you just use SP here?  I suspect the difference is less than the accuracy of most optodes.    

I've not been convinced GSW is great.  If your values depend on a spatial correction, one hopes they are very small...  

I think that means there needs to be info added to the attributes as to how this is processed.

callumrollo: I've had a look at the GSW code, it is indeed doing a geographical correction to SA, which is then used in calculation oxygen concentration. I've tested changing the lon and lat to various oceans (Atlantic, Pacific, Ross Sea, Grreenland Sea & Baltic). The changes to oxygen are way below the noise level of the optodes. Only the Baltic had changes I could see when zoomed right in. The others are near identical.

comp_zoom

Here's the mean oxygen concetration values. This is using the example-seaeplorer data

uncorr mean: 180.39353891647335
true mean: 145.7232575360533
trop_atl mean: 145.72527999183677
trop_sth_pac mean: 145.72592724849966
ross_sea mean: 145.72227471307164
baltic mean: 145.62095778937243

callumrollo: Here's a gist of the code I used in the above calculations

https://gist.github.com/callumrollo/d17cfef02b5830c23f2c21ebe46ecdd8

callumrollo: I'll update the attributes to reflect the processing

jklymak: OK, but my tendency would be to use the simplest processing possible, which means to use SP, not SA - if it really makes no practical difference, its easier to back out.

pyglider/utils.py
```diff
@@ -360,6 +360,25 @@ def nmea2deg(nmea):
...
+    ct = gsw.CT_from_t(sa, data['temperature'], data['pressure'])
+    o2_sol = gsw.O2sol(sa, ct, data['pressure'], data['longitude'], data['latitude'])
+    o2_sat = data['oxygen_concentration'] / gsw.O2sol(sa*0 + ref_sal, data['temperature'], data['pressure']*0, data['longitude'], data['latitude'])
+    data['oxygen_concentration'].values = o2_sat * o2_sol
+    data['oxygen_concentration'].attrs['comment'] = f'oxygen concentration corrected for salinity using gsw.02sol and ' \
```
jklymak, line 376: Is "comment" the right attribute here?  Do we have a "processing steps" attribute?  What does OG 1.0 say about detailing processing?  If we do use "comment" do we need to be careful to preserve any previous content of "comment" if it exists?

callumrollo: Good point. If we use comment, we should append rather than replacing.

As for OG 1.0, it looks like we should use ocygen_concentration_QC:RTQC_methodology 

https://github.com/OceanGlidersCommunity/OG-format-user-manual/blob/main/OG_Format.adoc#geophysical-variables

One should also link to the QC methodology. Guess I'd better read that oxygen SOP

callumrollo: Reference for salinity correction is to Gacia & Gordon 1992 https://aslopubs.onlinelibrary.wiley.com/doi/abs/10.4319/lo.1992.37.6.1307

And the argo oxygen processing manual

https://archimer.ifremer.fr/doc/00287/39795/

jklymak: That seems better than "comment'.  I'm not so concerned that the contents of this are correct, or even that the attribute name is the final one.  Just if we use "comment" we'd not want to lose previous/other comments, and that perhaps that is too general.  So I'm fine with whatever we use in here and we can always refine later.

richardsc:
Just to confirm (my comment/review might have gotten lost in the other conversation): Can we just use the gsw `gsw_o2sol_sp_pt()` instead of the TEOS one, since it's all the old formulae anyway? That removes the need to put longitude and latitude into the calculation at all.

callumrollo:
Great idea Clark! Sorry, I completely missed your suggestion earlier. I've implemented it now. I'll make the suggested change to how we record the processing step then hopefully we can merge this.

jklymak (approved):

Issue #20, opened 2021-10-25
convert to gsw


callumrollo:
Linked to Issue #15, the seawater package is deprecated. We should aim to migrate to gsw once tests are in place to ensure nothing is changed

callumrollo:
Resolved in #44

PR #44, opened 2021-12-09
Move to gsw

callumrollo:
This resolves Issue #20 

The move to gsw would remove a deprecated dependency and modernise the package. This is an essential step if we eventually want to get this on pypi/conda-forge.

Most of the changes went without a problem. There are a couple of hiccups: the depth produced by gsw differs very slightly (sub mm scale) from that in seawater. Consequently all tests against old data fail. To get past this, I have temporarily dropped the depth dimension before testing results against the test data, so we can focus on the variables that are actually changing. These are:

salinity
potential_density
density
potential_temperature
distance_over_ground

Of these, only distance_over_ground is substantially different. It increases by a factor of around 1.7


callumrollo:
depth

density


potential_density

potential_temperature

salinity

distance_over_ground

All these images are from the **example-seaexplorer-legato-flntu-arod-ad2cp** dataset

jklymak:
1.7 is the difference between km and nautical miles, so I think that is just a  setting.  We should be using km, I would argue.

callumrollo:
That was what I suspected. gsw outputs m by default. I divided by 1000 to get km. As the gsw distance is 1.7 * the seatwater distance, that would suggest seawater was outputting nautical miles. Which is odd as I though it should do km by default.

I'll look into it

callumrollo:
Are the changes in the other variables small enough to be acceptable? Are there further tests we should do before a switch to gsw?

jklymak:
I think we are testing that the process works, and that values don't change unexpectedly.  Here we expect a change, and indeed you saw a sizeable change in one variable.  So I'd say the tests are doing their job, and its OK to change the "truth" files once we are happy with them.

callumrollo:
Found the issue with distance. gsw.distance takes (lon, lat) where seawater.dist takes (lat, lon). Corrected now

distance_over_ground

callumrollo:
I'm happy with the small changes in the other variables. Shall I update the testcase data with gsw output?

jklymak:
Thats fine with me!  Thanks!

callumrollo:
Done :) please merge once the tests have passed

richardsc:
Late to this party but all looks good to me.

PR #45, opened 2021-12-15
catch invalid datetimes in AD2CP_TIME

callumrollo:
At the beginning of some missions, we get timestamps from AD2CP_TIME before it has synced it's clock. These are:

00000000 00:00:00

Passing one of these to pd.to_datetime to parse the dates throws an error

`ValueError: year 0 is out of range`

This addition checks for timestamps starting 000000 and replaces them with None, which pd.to_datetime then converts to NaT

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -111,6 +111,8 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True, min_samples_in
                                             dayfirst=True)
                     # If AD2CP data present, convert the timestamps to datetime type
                     if 'AD2CP_TIME' in out.columns:
+                        # Set datestamps with date 00000 to None to prevent datetime error
+                        out.loc[out.AD2CP_TIME.str[:6]=='000000','AD2CP_TIME'] = None
                         out['AD2CP_TIME'] = pd.to_datetime(out.AD2CP_TIME)
                     with out.to_xarray() as outx:
 


pyglider/seaexplorer.py
```diff
@@ -111,6 +111,8 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True, min_samples_in
                                             dayfirst=True)
                     # If AD2CP data present, convert the timestamps to datetime type
                     if 'AD2CP_TIME' in out.columns:
+                        # Set datestamps with date 00000 to None to prevent datetime error
+                        out.loc[out.AD2CP_TIME.str[:6]=='000000','AD2CP_TIME'] = None
```
jklymak, line 115: I missed that we had picked up a pandas dependency - is pandas really necessary here?  I'm also not clear why time is being saved as seconds since 1970 instead of datetime64 below?  I suspect this is historical in your adcp processing, but I'm not sure it is necessary.

callumrollo: We already had pandas in `raw_to_rawnc` to read in the csv input files. The data at this point in processing is a dataframe already so I thought it made sense to use

jklymak: ha, shows you what I remember about my own code.  OK, that seems fine.

callumrollo: the part with saving to seconds since 1970 also appears to be original

callumrollo: We could change it though, certainly

jklymak: No lets leave it for now.  Certainly not the issue here.

jklymak (approved):

pyglider/seaexplorer.py
callumrollo: We've all been there! the time could use a look at though. I guess that could be done as part of the the move to OG 1.0 once it's finalised

PR #46, opened 2021-12-15
changed all print to log.info or log.debug

callumrollo:
I'm automating processing of our seaexplorer data. The logging is very useful, but the print statements throw a lot of non-essential info into std out, which makes investigating errors a bit of a chore. 

As a suggestion, I have replaced all print statements with _log.info where they are useful, one line statements and _log.debug when they are larger, or appear designed more for troubleshooting than regular operation.

None of these suggestions are set in stone. What do you think?


jklymak (approved):
LGTM!  Thanks for doing this...

callumrollo:
Oops, should've run it first! logging library isn't happy with some of the messages. Will reformat them and add to the PR

PR #48, opened 2021-12-16
Migrate from seconds since 1970 to datetime64

callumrollo:


callumrollo:
This one way a pain. I've now refactored both seaeplorer and slocum processing the use datetime64 rather than seconds since 1970. There are a couple of consequences of this:

- we can now processes variables that are datetime, like ad2cp_time using pandas smart dateparser. See `example-seaexplorer-legato-flntu-arod-ad2cp`.
- As we are now saving time and other variables such as profile_time as datetimes rather than ints, we must remove the units from these before saving to netcdf, as xarray doesn't allow dtpe datetime64 dataarrays to have units. This is done automatically in the code, I've also removed the relevant yaml lines
- All the tests against data will fail, as the datetimes will all have changed by a few 100 nanoseconds

callumrollo:
I've manually checked the output and and 99 % there are no significant changes. I'll check again tomorrow and update the test data to have the new time data

callumrollo:
image
Alright. I'm pretty well convinced that the only changes are at the millisecond scale. Here's an overplot of heading. Can't even see the old data (blue) under the new (orange). Updating test data now>

The mean difference between variables is lower than numpy's detection level for most variables:

```
heading mean difference:    -3.709885731950635e-10 
pitch mean difference:    -1.1926459819733282e-10 
roll mean difference:    3.4685254668431753e-10 
conductivity mean difference:    0.0 
temperature mean difference:    0.0 
pressure mean difference:    0.0 
chlorophyll mean difference:    0.0 
cdom mean difference:    0.0 
backscatter_700 mean difference:    0.0 
oxygen_concentration mean difference:    -9.171969850285677e-10 
temperature_oxygen mean difference:    -5.8406612879480235e-12 
distance_over_ground mean difference:    -3.90087961932295e-12 
profile_index mean difference:    0.0 
profile_direction mean difference:    0.0 
salinity mean difference:    0.0 
potential_density mean difference:    0.0 
density mean difference:    0.0 
potential_temperature mean difference:    0.0 

```

jklymak (commented):
This seems fine.  However, do the resulting files have units for time?

callumrollo:
If loaded with xarray using drfault arguments, it recognises datetime and reads it as datetime64 with no units:
time.values
```
array(['2019-07-21T23:25:02.737000000', '2019-07-21T23:25:32.745000000',
       '2019-07-21T23:26:02.753000000', ...,
       '2019-07-23T21:24:46.390000000', '2019-07-23T21:25:16.396000000',
       '2019-07-23T21:25:46.402000000'], dtype='datetime64[ns]')
```
time.attrs
```
{'source': 'time',
 'long_name': 'Time',
 'standard_name': 'time',
 'axis': 'T',
 'observation_type': 'measured'}
```

If you set `decode_times=False` it reads the saved format, milliseconds since the time of the first sample:

time.values
```
array([        0,     30008,     60016, ..., 165583653, 165613659,
       165643665])
```
time.attrs
```
{'source': 'time',
 'long_name': 'Time',
 'standard_name': 'time',
 'axis': 'T',
 'observation_type': 'measured',
 'units': 'milliseconds since 2019-07-21 23:25:02.737000',
 'calendar': 'proleptic_gregorian'}
```
Thjis is a different calendar and units from those in the old yaml files, so I added checks to remove the units and calendar attributes, or xarray can't overwrite them and fails the nc write

callumrollo:
Here's a file with the new times:

https://callumrollo.com/nextcloud/index.php/s/D4DJ9iLowNH7DY2

jklymak:
Hmmm maybe that is why I did this in the first place. Because o don't think we can leave time with no units and be cf compliant?

callumrollo:
I think removing units is just what xarray does when it loads and interprets the data. If I load it with netCDF4, I get the units:

```
<class 'netCDF4._netCDF4.Variable'>
int64 time(time)
    source: time
    long_name: Time
    standard_name: time
    axis: T
    observation_type: measured
    units: milliseconds since 2019-07-21 23:25:02.737000
    calendar: proleptic_gregorian
unlimited dimensions: 
current shape = (5521,)
```

It includes the units and calendar, so I think it's still CF compliant?

callumrollo:
I get the same behaviour with the current netcdfs. Loading with xarray converts to datetime with no units, setting decode_times=False gives the units:
```
 {'source': 'time',
 'long_name': 'Time',
 'standard_name': 'time',
 'calendar': 'gregorian',
 'units': 'seconds since 1970-01-01T00:00:00Z',
 'axis': 'T',
 'observation_type': 'measured'}
 ```
 
 Just it's seconds since 1970 rather than milliseconds since first sample. Probably the cause of the nanosecond level differences in the results

**Comment by richardsc** [comment]
I can confirm looking directly at the netcdf that there are units for the time dimension:

```
1 dimensions:
time  Size:5521                                                                                           
source: time                                                                                          
long_name: Time                                                                                       
standard_name: time                                                                                   
axis: T                                                                                               
observation_type: measured                                                                            
units: milliseconds since 2019-07-21 23:25:02.737000                                                  
calendar: proleptic_gregorian                                                              
```

richardsc:
Relatedly, do we want the netCDF to have something sensible in terms of a date reference (e.g. POSIX)? Milliseconds since the first sample doesn't seem standard to me, though I haven't worked with a lot of similar netCDF data recently.

callumrollo:
xarray states that it defaults to the smallest unit of difference between sample points and the 'proleptic_gregorian' calendar  https://xarray.pydata.org/en/stable/user-guide/io.html#time-units

So that'll be why we're seeing this encoding. We can override the default behaviour when writing to though

https://github.com/pydata/xarray/issues/1324

This way we could use datetime internally, allowing us to load other variables that might be datetime, and use an encoding like seconds since 1970 at write time. I've not got much CF experience, but happy to implement whatever datetime encoding you judge most fit

jklymak:
I don't personally care because I use xarray but I think we should be cf compliant and og 1.0 compliant.  But I haven't had a chance yet to double check what that means yet 😀

callumrollo:
Looks like OG want seconds since 1970 https://github.com/OceanGlidersCommunity/OG-format-user-manual/blob/main/OG_Format.adoc#along-track-positioning-variables

I guess that'll bring us back to the original output format, just with internal workings based on datetime64. I'll work that into the code this eve

callumrollo:
OK the timestamps are now encoded like

{'source': 'time',
 'long_name': 'Time',
 'standard_name': 'time',
 'axis': 'T',
 'observation_type': 'measured',
 'units': 'seconds since 1970-01-01T00:00:00+00:00',
 'calendar': 'proleptic_gregorian'}
 
 That should satisfy CF 1.6 and OG 1.0
 
 @richardsc can you check the new file please? https://callumrollo.com/nextcloud/index.php/s/YbrK2nRSo72AxfG

richardsc:
Yup, looks good from the netCDF file!

callumrollo:
Nice. I'll merge this now

PR #50, opened 2022-01-28
DOC: improve explanation of web creation and include templates

jklymak:


jklymak:
When the tests pass I'll merge because we have a collaborator who wanted to try and make the web index pages and will need the templates.

PR #51, opened 2022-04-28
pyglider additions, adding argopandas as submodule

juliaputko:


jklymak:
I'm confused why we would add argo stuff to this repo?

juliaputko:
Hi Jody,

I am not completely sure if adding it as a submodule in this way is needed, but I am using the argo package to extract the argo trajectories for the home page map, and adding it as a subdirectory within pyglider is what worked best for me. 

Best, 
Julia Putko

hvdosser (changes requested):
Need to remove argo stuff and make a separate pull request for slocum.py and seaexplorer.py.

Issue #52, opened 2022-05-06
oxygen correction wrongly nans data in raw mode


callumrollo:
If the oxygen correction is applied to raw SeaExplorer data, the majority of oxygen data can be mistakenly converted to nan. This is due to a timestamps issue.

Oxygen data and CTD data are typically recorded at different times by the SeaExplorer, most CTD observations do not have an exactly corresponding oxygen observation, though they may be temporally very close. `glider_utils.oxy.oxygen_concentration_correction` applies a correction on the assumption that oxygen measurements have corresponding temperature and salinity measurements which are used to correct for the 0 salinity assumed by the AROD oxygen optode. This is often not the case. Where the oxygen data does not have corresponding temp/sal it is converted to nan.

This issue does not occur with sub data, as used in the tests, as these data are comprised of lines where each timestamp has a measurement for every variable. For this reason, the issue escaped detection by the tests.

PR #53, opened 2022-05-06
use nearest tempsal to correct oxygen

callumrollo:
This PR resolved Issue #52 

Rather than only using temperature and salinity from exactly matching timestamps, it used `pandas.merge_asof` to find the nearest ctd sample to use for the correction.

If no ctd measurement is found within the tolerance (provisionally set to 5 seconds) the oxygen value will be converted to nan.

### DIFF
--- a/pyglider/utils.py
+++ b/pyglider/utils.py
@@ -1,6 +1,8 @@
 import collections
+import datetime
 import xarray as xr
 import numpy as np
+import pandas as pd
 from scipy.signal import argrelextrema
 import gsw
 # import webcolors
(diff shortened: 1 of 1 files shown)


pyglider/utils.py
```diff
@@ -369,10 +371,17 @@ def oxygen_concentration_correction(data, ncvar):
         ref_sal = 0
     else:
         ref_sal = float(oxy_yaml['reference_salinity'])
+    # Find the closest temperature and salinity measurements to oxygen measurements with pandas timestamp matching
+    df_oxy = pd.DataFrame({"time": data.time, "oxygen_concentration": data.oxygen_concentration})
+    df_oxy = df_oxy[~np.isnan(df_oxy.oxygen_concentration)]
+    df_ctd = pd.DataFrame({"time": data.time, "salinity": data.salinity, "potential_temperature": data.potential_temperature})
+    df_ctd = df_ctd[~np.isnan(df_ctd.salinity)]
+    # Match the nearest timestamp within 5 seconds. Oxygen values with no temp/sal within 5 seconds will become nan
+    df_nearest = pd.merge_asof(df_oxy, df_ctd, on="time", direction="nearest", tolerance=datetime.timedelta(seconds=5))
```
jklymak, line 380: seems a little clunky to go back and forth from pandas like this.   Maybe worth the simplification, but is there no equivalent in xarray?

callumrollo: Good idea. I found a serviceable equivalent with [da.reindex](https://github.com/pydata/xarray/issues/3218)

I'll submit a different PR using that method

callumrollo:
Closed in preference of #54

PR #54, opened 2022-05-09
match timestamps of oxygen and ctd before correcting

### LINKED ISSUE #52: oxygen correction wrongly nans data in raw mode

callumrollo:
This PR resolves #52 

Rather than only using temperature and salinity from exactly matching timestamps, it used xarray da.reindex to find the nearest ctd sample to use for the correction.

### DIFF
--- a/pyglider/utils.py
+++ b/pyglider/utils.py
@@ -370,9 +370,13 @@ def oxygen_concentration_correction(data, ncvar):
     else:
         ref_sal = float(oxy_yaml['reference_salinity'])
     _log.info(f'Correcting oxygen using reference salinity {ref_sal} PSU')
-    o2_sol = gsw.O2sol_SP_pt(data['salinity'], data['potential_temperature'])
-    o2_sat = data['oxygen_concentration'] / gsw.O2sol_SP_pt(data['salinity']*0 + ref_sal, data['potential_temperature'])
-    data['oxygen_concentration'].values = o2_sat * o2_sol
+    ds_oxy = data.oxygen_concentration[~np.isnan(data.oxygen_concentration)]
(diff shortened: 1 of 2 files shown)


jklymak:
This seems like it should be OK?   I guess a big time gap would kind of suck, but that seems rare enough that this should work.  This is just to translate from saturation to concentration, right, so we can always reverse it easily enough?

callumrollo:
Yes it's reversible. The functions corrects for salinity using data from the CTD, rather than the constant salinity assumed by some oxygen optodes like the AROD. Big time gaps will be rare, as the CTD should always be sampling when the oxygen sensor is sampling. This function only runs if requested in the yml, so operators will be aware. 

I'll delete the other PR. Happy to merge this if you are?

jklymak:
Thanks!  BTW, hope your cruise went well.

callumrollo:
Had a great time thanks! Now making up for a two month break in my commit record :P

PR #56, opened 2022-05-11
Fixed error with time variable when writing netcdf file

hvdosser:
There was a problem in seaexplorer.py because of how 
xarray handles the encoding for the calendar in attrs on 
variable 'time'. To fix it, I replicated the code used to fix 
the same issue with the encoding for the units in attrs on 
variable 'time'. There may be other places this fix is needed.


jklymak:
That makes sense.  However, it would be nice if the yaml could somehow indicate the units in a way that xarray was happy with?

jklymak:
I do wonder if both these pops merit a warning that we are ignoring what is in the yaml....

PR #57, opened 2022-05-13
remove rogue 1970s

### LINKED ISSUE #55: 1970s from glider reboots
If a seaexplorer reboots mid-mission, a few 1970 datestamps will be produced before it gets a timefix from the GPS. This can mess up processing down the line, particularly as xarray assumes monotonically increasing datetime index for several operations. 

It would make sense to fix this as far upstream as possible. A crude but simple fix could be to drop any lines with datestamps before 1971 at the raw to rawnc stage. Or should it be left until later?

callumrollo:
This PR resolves #55 

If all timestamps are before or after 1970-01-01, no changes are made. If there are timestamps both before and after 1971-01-01, those before are dropped.

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -141,6 +141,26 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True, min_samples_in
     _log.info('All raw files converted to nc')
     return True
 
+
+def drop_rogue_1970(ds):
+    """
+    If dates greater than 1971, 1, 1 are observed, drop any dates before 1971-01-01, from the datset and return it.
+    This function removes 1970 timestamps caused by a SeaExplorer rebooting during a mission.
+    If all dates are < 1971-01-01, no action is taken
+
+    Parameters:
+        ds: xarray.DataSet
+            dataset to check for pre-1971 dates
+    Returns:
+        ds: xarray.DataSet
+    """
+    dt_1971 = pd.to_datetime(datetime.datetime(1971, 1, 1))
+    # If all dates before or after 1971-01-01, return the dataset
+    if (ds.time > dt_1971).all() or (ds.time < dt_1971).all():
+        return ds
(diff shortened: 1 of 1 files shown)


pyglider/seaexplorer.py
```diff
@@ -141,6 +141,26 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True, min_samples_in
...
+    This function removes 1970 timestamps caused by a SeaExplorer rebooting during a mission.
+    If all dates are < 1971-01-01, no action is taken
+
+    Parameters:
+        ds: xarray.DataSet
+            dataset to check for pre-1971 dates
+    Returns:
+        ds: xarray.DataSet
+    """
+    dt_1971 = pd.to_datetime(datetime.datetime(1971, 1, 1))
```
jklymak, line 157: Does np.datetime64("1971-01-01") work here?

pyglider/seaexplorer.py
```diff
@@ -141,6 +141,26 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True, min_samples_in
...
+    # If all dates before or after 1971-01-01, return the dataset
+    if (ds.time > dt_1971).all() or (ds.time < dt_1971).all():
```
jklymak, line 159: Is dropping them safe or should we give them better timestamps?   I think it's always a couple of bad times and the little bit of data is safe to drop (and usually not in the ocean), but just checking

pyglider/seaexplorer.py
callumrollo: It does. I've switched to using numpy

pyglider/seaexplorer.py
callumrollo: I just checked with the glider we had that rebooted. Around 0.015 % of the datapoints have 1970 timestamps, and the majority of those have no associated sensor data. I think we're safe to drop the lines. This also avoids the difficulty of interpolating timestamps.

jklymak (approved):

PR #59, opened 2022-05-23
allow alternative averaging strategies for gridded product

### LINKED ISSUE #58: gridding PAR - should log first
As recommend by Basiten Queste: PAR should be logged before gridding. This is sensible as PAR is an exponential variable. We wouldn't want to bias it when performing the mean while gridding.

This changed averaging strategy should be logged in the metadata of the variable. Maybe it should be controllable from the yaml too?

callumrollo:
Allow the user to specify averaging scheme used by scipy.stats.binned_statistic_2d when creating the gridded netCDF. Defaults to using mean as is currently set.

This resolved #58


callumrollo:
I am raising warnings when this functionality is called. 

The process for setting the geometric mean is a bit clumsy. Not sure how else to do it though

jklymak:
I don't think I'd raise a warning but I would include the method in the attrs of the variable.

jklymak (approved):

jklymak:
This looks cool.

PR #74, opened 2022-06-04
Mnt cleanup slocum

jklymak:
Pretty big cleanup to slocum.  Got rid of all the intermediary files.  Those were me trying to work around the lock=True problem for xarray.mfdataset.  Tested on Rosie, and on a full-resolution Walle data set.  Seems to work fine.


jklymak:
Tests were failing, but I think that is because of poor interpolation at that boundary between the divided files, and that this version is better.

jklymak:
Tests had to be reduced a bit from equality, I think due to roundoff issues.  Its too bad xarray doesn't have allclose...

PR #75, opened 2022-06-06
Tst rearrange tests

jklymak:
Use get_examples to get the examples from pyglider-example-data (https://github.com/c-proof/pyglider-example-data)


jklymak:
This breaks the examples in the docs, so still WIP

jklymak:
@callumrollo this removes the example data from the main repo and adds ways of downloading them from `pydata-example-data`

jklymak:
OK, I kept the examples the same.  The user should be installing this and running the examples from wherever they want, versus running from a development environment.  This is all angling towards a conda/pip recipe.

But I did add to the .gitignore because of course developers may decide to download the examples.  They will just download them somewhere different than the tests will download them.

callumrollo (approved):

PR #77, opened 2022-06-07
MNT: remove vendored bitstring

jklymak:
`pip install bitstring` works fine.


jklymak:
I'dll merge this as it just affects the slocums, though @callumrollo  if you want to discuss the merge policy, at some point we can.  I've basically been asking you to review anything that may touch seaexplorers.  At some point we should be more formal.

PR #82, opened 2022-06-15
DOC: document utils

jklymak:
Document the `utils.py` file.


jklymak:
@callumrollo For discussion:  This PR makes clear a pretty big problem with having the example data separate from the repo.    In particular, I changed some of the metadata in the processing.  So in order to make the tests pass I'd have to make a new version of the example data.  However, that is problematic because the new version hasn't been merged yet, and so _your_ tests would start to fail, at least until you merged the new changes.

I'm not sure what the fix is here.  We could make format changes have two accompanying PR requests.  The tests though would need some way of downloading the zip file for the correct branch of the example data.  That seems doable, but maybe fragile. 

I'm of a mind to put the example data back into the repo.  People who do pip-install won't get the example data so it doesn't take up too much space.  Developers can probably survive whatever load we put on them.

callumrollo:
Yes, I hadn't foreseen this. As the example datasets are small and don't get included in a pip install I don't see an issue with putting them back in the main repo

jklymak:
@callumrollo note the new less stringent test here.  I was locally getting failures without using all-close using the netcdf files you generated.  I think for our purposes allclose will be good enough, versus strict equality of the variables.
 
The test files had to be regenerated to allow for the attribute changes `get_profiles`->`get_profiles_new`.  note we had been using `get_profiles_new` for quite a while, but it was filling in the attribute incorrectly.

callumrollo:
I wasn't aware of all-close before. Looks good, should prevent spurious test failures

PR #85, opened 2022-06-15
TST: move examples back

jklymak:
Moves examples back to GitHub.


jklymak:
As a follow up, this needs - docs changed and to point the get_examples (which I think we should keep) to the proper place to make a zip file...

callumrollo (approved):
Looks good to me!

PR #89, opened 2022-06-16
remove plotting

jklymak:
This removes the plotting from pyglider, and from the example processing scripts.

Thinking about the a bit more, I'm of two minds.   The plotting is a pretty decent smoktest of the data.  Maybe we will add the c-proof/cproofutils as a testing dependency.


jklymak:
Self merging so I can get the tests working in #82

PR #97, opened 2022-06-21
FIX: get example data from cproof

jklymak:


jklymak:
@callumrollo Does this seem reasonable?  We can just have a static place to keep _some_ example data.  It need not be in sync with the test data, though we could make it so periodically.

PR #103, opened 2022-06-30
Update header check in Slocup.py to work with the new gliders

hvdosser:
Add two 'and' statements to the 'if' loops in the _check_diag_header
function in slocum.py, to account for the new header values for
data from the new Slocum G3 gliders.


jklymak:
Great Hayley (plus or minus getting the logic to work). I think we should add a small G3 file to the test data set so we can keep testing this, but that can wait until you have it working...

PR #107, opened 2022-07-15
Modified seaexplorer.py so it will run without .gli files present

hvdosser:


pyglider/seaexplorer.py
```diff
@@ -136,10 +136,9 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
                 _log.warning('Some files could not be parsed:')
                 for fn in badfiles:
                     _log.warning('%s', fn)
-            if not goodfiles:
-                _log.warning(f'No valid unprocessed seaexplorer files found in'
-                             f'{indir}')
-                return False
+    if not goodfiles:
```
jklymak, line 139: I guess this is fine.  I'd still prefer some explanation of what this fixes and why this is the correct fix.  Because this _seems_ like it should be fine.

pyglider/seaexplorer.py
```diff
@@ -99,7 +100,6 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
                 continue
 
             badfiles = []
-            goodfiles = []
```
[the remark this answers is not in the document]
hvdosser: I had been thinking we wanted each type of bad file logged separately, but if that's not the case then certainly it can get the same treatment.

pyglider/seaexplorer.py
hvdosser: Changing the location of the `if not goodfiles:` statement prevents the function from returning a value of `False`  before running the `pld1` for-loop. The `return False` was previously within the `for rawsub in ['raw', 'sub']:` for-loop, meaning that a value of `False` could be returned before any `pld1` files were processed. Moving the `if not goodfiles:` statement out of that for-loop ensures that we will process both `gld` and `pld1` files before the function returns anything.

jklymak (approved):

PR #109, opened 2022-08-21
Enh use dbdreader

jklymak:
This PR allows us to use https://dbdreader.readthedocs.io/ rather than the python-only slocum tools for the slocum gliders.  This is _much_ faster, and seems more robust as well.  

This probably needs more documentation.


jklymak:
@hvdosser self merging this.  Note the change in the default script for slocum gliders.

PR #115, opened 2022-09-20
ENH: use reindex_like to line up times

jklymak:
Instead of a for-loop, try re-indexing using `index_like`.


jklymak:
Tested on dfo-hal1002-20220914.  Works fine.

jklymak:
Checked on real-time processing and works

PR #118, opened 2022-10-02
FIX: refactor dbdreader code to just do one pass

jklymak:
dbdreader will sync other variables with a main variable.  Here make the main variable adjustable, and `sci_water_temp` by default.   Some troublesome files work better under this scheme than the old scheme of using time (the time stamp seemed corrupt in the bad files).  

The only issue here may be memory usage for large data sets?   There may need to be a way to read files in batches at some point, but hopefully this works for now.  If it doesn't another approach is to go back to the old way, and just sync this way.

### DIFF
--- a/pyglider/slocum.py
+++ b/pyglider/slocum.py
@@ -789,7 +789,7 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, *,
 
 
 def binary_to_timeseries(indir, cachedir, outdir, deploymentyaml, *,
-                         search='*.[D|E]BD',
+                         search='*.[D|E]BD', time_base='sci_water_temp',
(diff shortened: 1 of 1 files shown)


jklymak:
@hvdosser if you get a chance this week, can you run this version on a delayed-mode data set and make sure the results are comparable to the old processing?   I think some of the interpolation will be different, but...

jklymak:
@hvdosser did you try this branch yet?  If it's OK, please merge.  Thanks!

hvdosser:
Compared branch with main branch for delayed mode data for one mission. The L0 time series and grid files were very slightly smaller, and the processing ran more quickly. Data looked good, and metadata was unchanged. Merging.

PR #120, opened 2022-10-06
use polars for seaexplorer data file load

### LINKED ISSUE #36: Slow processing of large datasets

callumrollo:
This is a big PR. More work needed to check that it's not changing data as the methods are not the same as pandas. 

Testing on our datasets has shown a ~ 10 X speedup of processing large delayed mode datasets, and a ~5 X speedup with nrt data. I'll write some example code comparing the two methods along with more tests. This is almost certainly a sub-optimal implementation of polars, as I'm directly aping the existing flow for pandas.

Using polars also decreases disk usage of intermediate products by using parquet rather than .nc. It also has substantially lower memory usage than pandas, so should decrease overheads when processing large datasets.

This is designed to resolve #36

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -342,9 +342,16 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
                     coarsen_time = ncvar[name]['coarsen']
                     coarse_ints = np.arange(0, len(sensor)/coarsen_time, 1/coarsen_time).astype(int)
                     sensor_sub = sensor.with_columns(pl.lit(coarse_ints).alias("coarse_ints"))
-                    sensor_sub_coarse = sensor_sub.groupby('coarse_ints').median().sort(by="time")
-                    val2 = sensor_sub_coarse.select(sensorname).to_numpy()[:, 0]
-                    val = _interp_gli_to_pld(sensor_sub_coarse, sensor, val2, indctd)
+                    sensor_sub_grouped = sensor_sub.with_column(
+                        pl.col('time').to_physical()
+                    ).groupby(
+                        by=pl.col('coarse_ints'),
+                        maintain_order=True
+                    ).mean().with_column(
+                        pl.col('time').cast(pl.Datetime('ms'))
+                    )[:-1, :]
+                    val2 = sensor_sub_grouped.select(sensorname).to_numpy()[:, 0]
+                    val = _interp_gli_to_pld(sensor_sub_grouped, sensor, val2, indctd)
                 val = val[indctd]
 
                 ncvar['method'] = 'linear fill'
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -125,7 +125,11 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
                 _log.info(f'{f} to {fnout}')
                 if not incremental or _needsupdating(ftype, f, fnout):
                     _log.info(f'Doing: {f} to {fnout}')
-                    out = pl.read_csv(f, sep=';')
+                    try:
+                        out = pl.read_csv(f, sep=';')
+                    except:
+                        _log.warning(f'Could not read {f}')
+                        continue
                     if "Timestamp" in out.columns:
                         out = out.with_column(
                             pl.col("Timestamp").str.strptime(pl.Datetime, fmt="%d/%m/%Y %H:%M:%S"))
@@ -134,6 +138,9 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
                         out = out.with_column(
                             pl.col("PLD_REALTIMECLOCK").str.strptime(pl.Datetime, fmt="%d/%m/%Y %H:%M:%S.%3f"))
                         out = out.rename({"PLD_REALTIMECLOCK": "time"})
+                    for col_name in out.columns:
+                        if "time" not in col_name.lower():
+                            out = out.with_column(pl.col(col_name).cast(pl.Float64))
                     # If AD2CP data present, convert timestamps to datetime
                     if 'AD2CP_TIME' in out.columns:
                         # Set datestamps with date 00000 to None
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -350,10 +350,12 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
                 _log.debug('sensorname %s', sensorname)
                 val = convert(sensor.select(sensorname).to_numpy()[:, 0])
                 if 'coarsen' in ncvar[name]:
-                    # smooth oxygen data as originally perscribed
+                    # coarsen oxygen data as originally perscribed
                     coarsen_time = ncvar[name]['coarsen']
+                    # create a boolean mask of coarsened timesteps. Use this mask to create an array of samples to keep
                     coarse_ints = np.arange(0, len(sensor) / coarsen_time, 1 / coarsen_time).astype(int)
                     sensor_sub = sensor.with_columns(pl.lit(coarse_ints).alias("coarse_ints"))
+                    # Subsample the variable data keeping only the samples from the coarsened timeseries
                     sensor_sub_grouped = sensor_sub.with_column(
                         pl.col('time').to_physical()
                     ).groupby(
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -132,7 +132,7 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
                         out = out.rename({"Timestamp": "time"})
                     else:
                         out = out.with_column(
-                            pl.col("PLD_REALTIMECLOCK").str.strptime(pl.Datetime, fmt="%d/%m/%Y %H:%M:%S.%f"))
+                            pl.col("PLD_REALTIMECLOCK").str.strptime(pl.Datetime, fmt="%d/%m/%Y %H:%M:%S.%3f"))
                         out = out.rename({"PLD_REALTIMECLOCK": "time"})
                     # If AD2CP data present, convert timestamps to datetime
                     if 'AD2CP_TIME' in out.columns:
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -25,7 +25,7 @@ def _outputname(f, outdir):
     for ff in fns:
         fnout += ff.lower() + '.'
     filenum = int(fns[4])
-    return outdir + fnout + 'nc', filenum
+    return outdir + fnout + 'parquet', filenum
 
 
 def _needsupdating(ftype, fin, fout):
@@ -149,11 +149,11 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
 
                     if ftype == 'gli':
                         out = out.with_columns([(pl.col("NavState") * 0 + int(filenum)).alias("fnum")])
-                        out.write_parquet(fnout[:-3] + '.parquet')
+                        out.write_parquet(fnout)
                         goodfiles.append(f)
                     else:
                         if out.select("time").shape[0] > min_samples_in_file:
-                            out.write_parquet(fnout[:-3] + '.parquet')
+                            out.write_parquet(fnout)
                             goodfiles.append(f)
                         else:
                             _log.warning('Number of sensor data points'
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -252,9 +252,15 @@ def merge_rawnc(indir, outdir, deploymentyaml, incremental=False, kind='raw'):
 
 def _interp_gli_to_pld(gli, ds, val, indctd):
     gli_ind = ~np.isnan(val)
-    valout = np.interp(ds['time'],
-                       gli.filter(gli_ind)["time"],
-                       val[gli_ind])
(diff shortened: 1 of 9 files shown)


callumrollo:
This is failing some tests at the moment. Some of those are expected (they use intermediate ncs which are now parquet files) others look like some errors with timestamp parsing and some other issues. I'll work on these

jklymak:
I wonder if this needs to be its own method?  I'd need to be convinced that parquet files are a better intermediate format than netcdf.  We need to be able to look at the intermediate files to see what is in the raw data - what is the workflow for parquet files?  Just polars?  I don't think xarray handles them. Does the efficiency go away if you use netcdf and polars?

We recently had this issue with the raw Alseamar files and solved it by subsampling the redundant information and reduced the raw files down from 10Mb each to 40kb

callumrollo:
These are good points, thanks Jody. We've tried subsampling the raw alseamar files, but with the 16 Hz legato running for 3 week missions, we still end up with huge datasets and the load/merge step is taking several hours per dataset.

I'm starting to profile the code now. polars makes time savings over pandas when loading and merging the data. parquet files are quicker and more storage efficient than ncs to write and read, but they do need to be readable as intermediate products. This is achieved in a very similar way to pandas:

```
>>> df = pl.read_parquet("sea63_35_nrt_rawnc/sea063.0035.gli.sub.0142.parquet")
>>> df
shape: (107, 23)
┌─────────────────────┬──────────┬───────────────┬─────────┬─────┬────────┬─────────┬──────────┬──────┐
│ time                ┆ NavState ┆ SecurityLevel ┆ Heading ┆ ... ┆ AngPos ┆ Voltage ┆ Altitude ┆ fnum │
│ ---                 ┆ ---      ┆ ---           ┆ ---     ┆     ┆ ---    ┆ ---     ┆ ---      ┆ ---  │
│ datetime[μs]        ┆ i64      ┆ i64           ┆ f64     ┆     ┆ f64    ┆ f64     ┆ f64      ┆ i64  │
╞═════════════════════╪══════════╪═══════════════╪═════════╪═════╪════════╪═════════╪══════════╪══════╡
│ 2022-03-01 23:50:06 ┆ 117      ┆ 0             ┆ 152.82  ┆ ... ┆ -5.0   ┆ 28.5    ┆ -1.0     ┆ 142  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ 2022-03-01 23:50:11 ┆ 110      ┆ 0             ┆ 154.58  ┆ ... ┆ 0.5    ┆ 28.5    ┆ -1.0     ┆ 142  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ 2022-03-01 23:50:21 ┆ 110      ┆ 0             ┆ 151.92  ┆ ... ┆ 0.5    ┆ 28.4    ┆ -1.0     ┆ 142  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ 2022-03-01 23:50:31 ┆ 110      ┆ 0             ┆ 156.13  ┆ ... ┆ 0.5    ┆ 28.4    ┆ -1.0     ┆ 142  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ ...                 ┆ ...      ┆ ...           ┆ ...     ┆ ... ┆ ...    ┆ ...     ┆ ...      ┆ ...  │

```

to me this makes more sense, as at the load_merge stage there is no metadata to make the bulkier netcdfs necessary for these intermediate table-like files.

I'm working on a full demo at the moment. I'd like to make use of this speedup, but if it's incompatible with existing workflows I'll make it a separate function or just keep it for internal use at VOTO

callumrollo:
I've put together a rough notebook profiling the polars implementation against the current main branch. It shows the performance differences using sub and raw data as well as showing what the intermediate parquet products look like when loaded in polars

https://github.com/callumrollo/pyglider_profile/blob/main/profile_pyglider.ipynb

It's pretty ugly, but shows the main points. I can work to make it portable//repeatable if desired

jklymak:
OK 5-10x is a big deal if you have 24 Hz data.  If you do this, can you update the docs to explain what a parquet file is and a quick how-to on opening them?  I think we can assume pandas and xarray knowledge (we are writing to netcdf), but parquet probably needs a paragraph of explanation.  

Final question would be if parquet and pandas have any packaging issues?  Does parquet work with a `conda install pandas` on the big three OS's?  (Can't be worse than netcdf ;-)

truedichotomy:
Would Zarr be a possible format option for pyglider?

https://medium.com/pangeo/cloud-performant-reading-of-netcdf4-hdf5-data-using-the-zarr-library-1a95c5c92314

https://www.azavea.com/blog/2022/09/22/benchmarking-zarr-and-parquet-data-retrieval-using-the-national-water-model-nwm-in-a-cloud-native-environment/

callumrollo:
I've had a look and it turns out pandas is able to read and write parquet files, so user shouldn't ever need to interact with polars. I'll add a paragraph to the docs explaining it. 

I've not encountered problems adding polars to the build so far. I'll test it out on a windows machine today to be sure though. Polars has mature builds on PyPI and conda-forge

callumrollo:
No issues making conda environments in pandas. I'm testing it on a range of our SeaExplorer datasets now. I'll mark ready for review once I'm satisfied it's performing well in production

callumrollo:
I've tested this pretty extensively on our datasets now. I think it's good to go. @jklymak do you have any other requested changes?

jklymak:
Give me a day or two to look it over.  Probably is all good.

pyglider/seaexplorer.py
```diff
@@ -125,49 +125,59 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
                 _log.info(f'{f} to {fnout}')
                 if not incremental or _needsupdating(ftype, f, fnout):
                     _log.info(f'Doing: {f} to {fnout}')
-                    out = pd.read_csv(f, header=0, delimiter=';',
-                                      parse_dates=True, index_col=0,
-                                      dayfirst=True)
+                    try:
+                        out = pl.read_csv(f, sep=';')
+                    except:
+                        _log.warning(f'Could not read {f}')
```
jklymak, line 131: Do we need to add the `badfiles` here?

pyglider/seaexplorer.py
```diff
@@ -125,49 +125,59 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
...
+                        continue
+                    if "Timestamp" in out.columns:
```
jklymak, line 133: Can this get a comment?  Why do we need this check?

pyglider/seaexplorer.py
```diff
@@ -125,49 +125,59 @@ def raw_to_rawnc(indir, outdir, deploymentyaml, incremental=True,
...
-
-                    with out.to_xarray() as outx:
-                        key = list(outx.coords.keys())[0]
-                        outx = outx.rename({key: 'time'})
-                        outx['fnum'] = ('time',
-                                        int(filenum)*np.ones(len(outx['time'])))
-                        if ftype == 'gli':
-                            outx.to_netcdf(fnout[:-3] + '.nc', 'w')
+                    if rawsub == 'raw' and dropna_subset is not None:
+                        out = out.with_column(out.select(pl.col(dropna_subset).is_null().cast(pl.Int64))
```
jklymak, line 152: This needs a comment, and if it doesn't cause a slowdown or extra writes maybe would benefit from unpacking into separate calls.   Looks like you are dropping repeats, but its too many steps in one line for me to follow without spending too much time ;-)

pyglider/seaexplorer.py
```diff
@@ -180,11 +190,15 @@ def drop_rogue_1970(ds):
...
+    pre_1971 = df.filter(pl.col("time") < dt_1971)
+    if len(pre_1971) == len(df):
+        return pre_1971
+    post_1971 = df.filter(pl.col("time") > dt_1971)
+    if len(post_1971) == len(df):
+        return post_1971
+    return df.filter(pl.col("time") > dt_1971)
 
 
 def merge_rawnc(indir, outdir, deploymentyaml, incremental=False, kind='raw'):
```
jklymak, line 204: We should probably rename this?

jklymak (commented):
This mostly looks good, just a couple of style things, and we probably need to change a top-level name.  

I think we need to decide if we are used enough yet to need a deprecation cycle for removing things.   I think we are probably fine to do this now, but at some point we can't change a bunch of underlying code that changes the output file type(!) without a deprecation cycle.  In which case, code like this would be better as a new API and the old API kept around for users who need it, with a deprecation warning.

pyglider/seaexplorer.py
callumrollo: badfiles added

pyglider/seaexplorer.py
callumrollo: I've added a comment on this. It's a bit convoluted looking, but it's just the polars equivalent of pandas.dropna. I didn't want to change the functionality of the dropna option in this PR. We can factor it out as a separate PR though?

pyglider/seaexplorer.py
callumrollo: I've renamed this to the more descriptive `drop_pre_1971_samples`

callumrollo:
I've added some more comments and used badfiles if a file fails the load. I have no idea how these changes have caused a couple of the slocum tests to start failing

callumrollo:
OK I think we're good to go now. @jklymak has this met your requested changes? In future I'll make these kind of changes as part of a deprecation cycle as you suggest

pyglider/seaexplorer.py
jklymak: I meant rename merege_rawnc....  Though that screws with older scripts, but I'd rename this merge_parquet or whatever, and then alias merge_rawnc so old scripts work

pyglider/seaexplorer.py
```diff
@@ -317,35 +332,43 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
...
                 if 'coarsen' in ncvar[name]:
                     # smooth oxygen data as originally perscribed
                     coarsen_time = ncvar[name]['coarsen']
-                    sensor_sub = sensor.coarsen(time=coarsen_time,
-                                                boundary='trim').mean()
-                    val2 = sensor_sub[sensorname]
-                    val = _interp_gli_to_pld(sensor_sub, sensor, val2, indctd)
+                    coarse_ints = np.arange(0, len(sensor)/coarsen_time, 1/coarsen_time).astype(int)
+                    sensor_sub = sensor.with_columns(pl.lit(coarse_ints).alias("coarse_ints"))
+                    sensor_sub_grouped = sensor_sub.with_column(
```
jklymak, line 357: This block needs a comment as well...  It says "smooth" oxygen above, but I'm not following this danse with `coarse_ints` etc.

jklymak (commented):
Looks close, just a couple more places to clarify the code.

pyglider/seaexplorer.py
callumrollo: Ah that makes more sense! I've renamed to merge_parquet and added `merge_rawnc = merge_parquet` so hopefully older scripts will still work. I've also added more comments to the variables coarsening

jklymak:
@callumrollo feel free to self merge, perhaps squash commits that you don't need, or just squash on merge. Thanks! we are about to try this out on some data sets that have been giving us problems.  ping @hvdosser

richardsc:
We have at least one dataset that has 16Hz legato data (plus 1Hz GPCTD), which would probably benefit from this change. Pinging @clayton to try this out on that mission.

jklymak:
@richardsc @clayton This is now merged, and on master.  It's _not_ yet released, so you will need to install from the development branch.   Let us know if that is problematic.   If it manages to get through a few more glider setups we should release as well as the changes to the slocum processing.

Issue #111, opened 2022-09-07
MNT: Factor out profile computation...


jklymak:
Currently, `binary_to_timeseries` or `raw_to_timeseries` (Sea explorer) both have the "profile" finding logic in them.  Recent experience has shown that this step is often somewhat flaky, and causes the raw time series process to not complete.

I'd propose we break the "profile" logic into its own routine and ask users to call it after making the time series to add profile info.  That will also allow more modularity with the "best" profile logic, which I find depends on the sampling frequency of the pressure sensor.  It would also allow for taking advantage of any metadata sent up by the glider as well.  So now users would do:

```
    outname = slocum.raw_to_timeseries(
        rawdir, l1tsdir, deploymentyaml)
    outname = ncprocess.get_profiles(outname, ...)
```

PR #124, opened 2022-10-21
start refactor of get_profiles

callumrollo:
This will resolve Issue #111 

Failing a few tests atm


callumrollo:
@jklymak is this what you had in mind to resolve #111? `get_profiles` is now a separate function that works on the timeseries netCDF created by `raw_to_timeseries`. This should enable more flexible definition of profile_index before splitting into profiles/gridding.

Recently alseamar changed the way that dives are defined on the seaexplorer to a much more sensible method. It now starts a new dive file after the first GPS fix at surface, so we should be able to use a much simpler method to define profiles. Having factored out `get_profiles`  this can be written as a seperate function for end users to call.

callumrollo:
@jklymak I've merged in the latest changes to main and fixed this PR for factoring out get_profiles, if that's still a goal.

The commit history is a bit of a mess after getting stale, so likely better to squash and merge

jklymak:
I've not looked at this carefully yet, but 1) it should remain back-compatible so the raw-to-timeseries methods could still make the profile extraction work. 2) I think the profile finding already is a separate method operating on xarrays?  So the to-do would be to make that function easier to access, make/lightly deprecate doing it in the raw-to-timeseries, and change the examples to use the external interface.

callumrollo:
I think this will require a bit more planning, especially as any change to this will need end users to update their processing scripts. I'll close this PR for now and design something that is more robust. I'm wondering if a class based approach might work better

PR #127, opened 2022-11-08
Add attributes to the time variable in netcdf files

hvdosser:
Changed two lines in slocum.py to make sure that the attributes for the time variable are not getting lost or being specified twice. Attributes are from the metadata yaml file.


jklymak:
Thanks @hvdosser

PR #129, opened 2022-11-23
Fix time variable attributes for IOOS metadata

hvdosser:
Modified the attributes for the time variable in ncprocess.py to be compliant with IOOS requirements, including the calendar, units, and FillValue used.


jklymak:
I think the errors in the tests are from hard-coded tests that you will just need to change in the test files.

hvdosser:
@jklymak, am I clear to merge?

PR #132, opened 2022-12-07
Add codecov

jklymak:


jklymak:
This works.  Not quite sure if I need to do something more to get codecov reports to link here.  I'll merge and we can keep playing with it.

PR #136, opened 2022-12-07
TST: test codecove again; Do not merge

jklymak:
Another test, do not merge.


jklymak:
So, basically code coverage now works.  If we make a PR, test coverage should not go down, and ideally should go up.  We should aim for 100% test coverage at some point, but no rush.

Issue #135, opened 2022-12-07
sea explorer time base fallbacks don't work with parquet 


jklymak:
https://github.com/c-proof/pyglider/blob/67c6acdf2b37926758090c273caa971930ccd1a2/pyglider/seaexplorer.py#L323

These lines error out if timebase is not specified in the yaml.  I guess we could enforce timebase in the yaml, but this code should be fixed or removed.

callumrollo:
resolved in #139

Issue #128, opened 2022-11-16
SeaExplorer delayed mode time series data loss


hvdosser:
In testing the new code for this pull request, I found an issue with processing the delayed mode SeaExplorer time series data for missions for which certain sensors (oxygen in this case) are severely oversampled. These missions end up with delayed mode data files that contain fewer actual (non-nan) data points than the realtime files. In other words, we are losing data during the processing. 

Currently, the `dropna` function is used to remove the oversampled oxygen data when converting the raw data. The `dropna` function is working correctly, however note that the resulting data has many nan values in it, for both the CTD and optics. These nan values will often not co-occur. 

I think the problem in the processing is caused by using the `GPCTD_TEMPERATURE` as the default time base in `seaexplorer.py`. This variable contains nan values that are not all co-located with the nan values in the oxygen and optical variables. It's desirable to use the CTD as the time base, but we may need to do some interpolation to avoid losing data when the other variables are mapped onto this base.

callumrollo:
We have had similar issues with our delayed mode datasets in the past. To get around it, we use NAV_LATITUDE as the timebase in all of our datasets, as this has a non-nan value for every line. Here's an example yml for one of our gliders with a GPCTD. https://github.com/voto-ocean-knowledge/deployment-yaml/blob/master/mission_yaml/SEA70_M15.yml

Would this solution work for you?

This also raises the broader point that perhaps we should have a test with delayed mode data, as currently all tests are run with nrt data

callumrollo:
We use NAV_LATITUDE as timebase in conjunction with keep_variables that account for each of the sensors. In this way, all data pass the `dropna` function, but rows with no data from any of the sensors are dropped by the `'keep_variables' in ncvar` block.

richardsc:
I'm not sure if this is related, but I had previously noticed that the timestamps recorded for sensors that are known to be outputting at 1Hz (according to the sensor) are never exactly 1Hz, likely due to the time that they arrive at the payload computer -- in other words, timing information from the sensor is not recorded by the PLD but it assigns it's own. The result is that two different sensors sampling at the same rate (e.g. a CTD and an optical O2 sensor) end up with times that are different by microseconds from one sample to the next, and even though we expect them to be "simultaneous", in terms of the recorded time stamps, they aren't.

jklymak:
[...]

PR #137, opened 2022-12-08
fix non-specified timebase defaults to work with parquet

callumrollo:
this fixes the failure flagged in #135. I neglected to update the timebase fallbacks when changing from pandas to parquet


jklymak:
Oh weird - I had a comment, but must have erased it. 

1. Can you rebase on main so we get the codecov? 
2. I wonder if we should allow this fallback or just force users to specify timebase?

callumrollo:
With the changes proposed in #128 I agree it would probably be better to force users to specify a timebase. That way, the user will know which sensor/variable they are using to align other timestamps with. I'll open another PR removing the fallbacks and returning an error message if timebase is not specified.

Not sure why codecov didn't run for this, it's up to date with c-proof/main

PR #139, opened 2022-12-09
raise error if no valid timebase specified

callumrollo:
alternative solution to #137. Forces user to specify a timebase


callumrollo:
This commit adds tests for a timebase that doesn't match any of the variables and for if the timebase has been dropped from the yaml. In both cases, we check the error string for the expected error message.

callumrollo:
@jklymak is this ready to merge with the added test?

jklymak (approved):

jklymak:
Note I think that we should force timebase for slocums as well.

PR #140, opened 2022-12-15
interpolate over nans of pld sensors

callumrollo:
This will fix Issue #128 

This is a first draft. If `interpolate` is set to True in `raw_to_timeseries`, nans from variables will be interpolated over as the sensors are aligned with the timestamps of the variable designated as the timebase.

This should be tested with some delayed mode data, as the interpolation has no effect on nrt data, as these typically have no nans. @hvdosser @jklymak, do you have a delayed mode dataset you would like to use to test this method?


callumrollo:
We can make the interpolate kwarg True by default, or control this behavior from the yaml

hvdosser:
Hi @callumrollo, I'd recommend testing on this delayed mode dataset: https://cproof.uvic.ca/gliderdata/deployments/dfo-bb046/dfo-bb046-20200908/delayed_raw/

callumrollo:
Thanks @hvdosser. Can you send me a yaml file for the dataset you linked? I will then incorporate the nan interpolation with control from the yaml file.

Sorry for the delay on this, just back from vacation

hvdosser:
Hi @callumrollo, no worries, thanks for figuring this out! Here's the link to the yaml: https://cproof.uvic.ca/gliderdata/deployments/dfo-bb046/dfo-bb046-20200908/deployment.yml

callumrollo:
@hvdosser do these tests look sufficient? I've added some of the data you provided and tested it in default mode and with nan interpolation enabled. We're using GPCTD as the timebase.The nan interpolation adds non-nan values for the other sensors (oxygen and cdom) where they did not sample at the same time as the CTD.

Getting a nice increase isn test coverage from this too :)

jklymak:
Ping @hvdosser for a review on this

jklymak:
@hvdosser if you had a chance to make sure this works it would be appreciated...

callumrollo:
@hvdosser do you have a timeframe for reviewing this? It's quite a substantial change, so I'd like to get it merged before addressing other Issues

hvdosser:
Hi @callumrollo and @jklymak , sorry for the delay on this. I'll have time on Monday Feb 13th to review and merge.

tests/test_pyglider.py
```diff
@@ -57,6 +58,98 @@ def test_example_seaexplorer_metadata():
...
+    test_data_raw.attrs.pop('date_issued')
+    assert output_raw.attrs == test_data_raw.attrs
+
+
+# Test nan interpolation on raw data.
+
+# Test that interpolation over nans in raw data results in a greater or equal number of non-nan values
+with open(deploymentyaml_raw) as fin:
+    deployment_raw = yaml.safe_load(fin)
+interp_yaml = str(example_dir / 'example-seaexplorer-raw/deploymentRealtimeInterp.yml')
```
hvdosser, line 130: It won't affect the results of the tests, but the naming convention would more accurately be `deploymentDelayedInterp.yml`

tests/test_pyglider.py
```diff
@@ -57,6 +58,98 @@ def test_example_seaexplorer_metadata():
...
+deployment_raw['netcdf_variables']["interpolate"] = True
+with open(interp_yaml, "w") as fout:
+    yaml.dump(deployment_raw, fout)
+l0tsdir_interp_raw = str(example_dir / 'example-seaexplorer-raw/L0-timeseries-test-interp/') + '/'
```
hvdosser, line 134: I think this will overwrite the results of the test for the subsampled interpolated data above, which isn't a major issue, but it would be nice to have a separate directory for the delayed-mode interpolated data in case we want to manually look at the resulting files.

pyglider/seaexplorer.py
```diff
@@ -366,7 +370,21 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
...
+                    if val.dtype == "<M8[us]":
+                        # for datetime, must convert to numerical, interpolate, then convert back
+                        us_since_1970 = (var_non_nan - np.datetime64("1970-01-01")).astype(int)
+                        val_int = np.interp(time_timebase.astype(float), time_var.astype(float), us_since_1970)
+                        val_us = val_int.astype("timedelta64[us]")
+                        val = np.datetime64("1970-01-01") + val_us
+                    else:
+                        val = np.interp(time_timebase.astype(float), time_var.astype(float), var_non_nan)
+                else:
+                    val = val[indctd]
```
hvdosser, line 387: This may betray my ignorance with regards to testing, but why wouldn't line 387 here be covered by the tests? One of your tests covers the raw data with `interpolate=False`, which should include line 387, shouldn't it?

hvdosser (commented):
Looks good overall, and the new tests are great! I added a few very minor comments. As a broader comment for @jklymak, we'll likely want to update our example yaml files and documentation with some information about including `interpolate` for the SeaExplorers, since this is a pretty major correction to the delayed-mode processing. Thanks for your efforts on this @callumrollo, and apologies again for the delay in review.

pyglider/seaexplorer.py
callumrollo: It certainly should do. Not sure why codecov doesn't pick that up

tests/test_pyglider.py
callumrollo: It uses the same raw directory `rawnc` as the run on raw data above, as the interpolation step comes after the `raw_to_raw_nc` and `merge_parquet` functions. This prevents unnecessary re-runs of the code. It writes results out to a new directory `l0tsdir_interp_raw` which is different to `l0ts_dir` and `l0tsdir_interp` so that they can be compared with the results obtained not using the interpolation functionality. 

I appreciate the naming of these is quite confusing. I think all the examples could benefit from a cleanup/refactor in a future PR

tests/test_pyglider.py
callumrollo: Good point. I've changed the yaml name accordingly

callumrollo:
I've merged in the latest changes from main. Is this good to go @hvdosser?

hvdosser:
@callumrollo and @jklymak, I'll merge this pull request, but I do think we need to consider what happens when we have large gaps in the data and we interpolate over them. Particularly for sensors like the ECOpuck, which only samples a small portion of the water column. Am I correct in thinking that setting `interpolate=True` will result in new interpolated points everywhere the CTD has data? If that's the case, we'll need to figure something out to deal with these larger gaps before we can move forward.

jklymak:
OK, I _think_ xarray provides this functionality:

https://docs.xarray.dev/en/stable/generated/xarray.DataArray.interpolate_na.html#xarray-dataarray-interpolate-na

note the *max_gap* parameter, which should be perfect for this.  Just set to some reasonable value (that probably changes depending on the sampling frequency) and big gaps would remain NaN...

callumrollo:
Hi @hvdosser. The potential problem with slow sampling sensors was raised in the original discussion of Issue #128. Linear interpolation will result in an apparent measurement at every timestamp of the specified timebase. using `keep_vars` will ensure these measurements are retained. An alternative would be to implement a max_gap as Jody suggests, or to have the option of nearest neighbout interpolation of some kind. Either way, I feel this is best addressed in a separate PR, preferably with the use of a dataset which demonstrates the issue you want to solve as test data

PR #141, opened 2023-01-05
Fixed time_profile variable units to be IOOS compliant.

hvdosser:
The variable time_profile had units that were e.g., 'seconds from 2022-12-01' but with the time changing for each profile. Code has been modified so the
correct units will show up for all profiles. This
fix also solved an issue where the value of time_profile was showing up as zero for some reason.


jklymak:
Can we add a test for this.  e.g. test that the times are what we want when read back in?

hvdosser:
I don't think we have any tests for the profile netcdf files right now, or for the gridded files. Do you want me to write one?

Issue #123, opened 2022-10-21
Legato error values in pressure converted to depth


callumrollo:
The legato pressure sensor is used to calculate glider depth. When the sensor is switched off/malfunctioning, the seaexplorer reports a value of 9999 for pressure. This is converted to a depth of 9690 m by `utils.get_glider_depth`. While 9999 is a fairly obvious error value, 9690 is not.

This function has a filter step for "good" pressures which it later uses to interpolate over the bad depths. Should we exclude pressure ratings greater than the maximum pressure from the "good" subset of pressures used to calculate depth perhaps?

callumrollo:
Proposal: We should catch 9999 data points and flag them. Perhaps convert them to nan? Otherwise they can be processed to other numbers that are not obvious error codes

callumrollo:
Resolved in #143

PR #143, opened 2023-01-16
replace 9999 data values with nan

callumrollo:
This PR will fix Issue #123 where 9999 error values from the legato are interpreted as valid pressures. All 9999 are replaced with nan during `raw_to_timeseries`.

Need to add tests.


jklymak (approved):
Looks good to me!

PR #149, opened 2023-04-14
Fix for slocums, to interpolate over gaps smaller than 'maxgap'

hvdosser:
A new function `find_gaps` has been added to `utils.py`, which identified gaps in scientific variables, and fills any gaps longer than the user-specified parameter `maxgap` with nans.


pyglider/slocum.py
```diff
@@ -884,6 +884,16 @@ def binary_to_timeseries(indir, cachedir, outdir, deploymentyaml, *,
...
             _log.debug('Sci sensorname %s', sensorname)
             val = data[nn]
+
+            # interpolate only over those gaps that are smaller than 'maxgap'
+            _t, _ = dbd.get(ncvar[name]['source'])
+
+            try:
+                tg_ind = utils.find_gaps(_t,time,maxgap) 
+                val[tg_ind] = np.nan            
+            except ValueError:
```
jklymak, line 894: Did you hit this exception?  Why?  A print statement probably is wrong.  Either warm or log.

hvdosser: I never hit the exception, and the print statement should've been removed after testing. I've done that now. Thanks!

jklymak (approved):

PR #150, opened 2023-04-21
Fix for SeaExplorer code to avoid interpolating over large gaps

hvdosser:
This fix uses the same `utils.findgaps` function as the Slocum code, and only interpolates over gaps that are larger than the user-specified parameter `maxgap`.

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -388,9 +388,9 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
                     else:
                         val = np.interp(time_timebase.astype(float), time_var.astype(float), var_non_nan)
 
-                        # interpolate only over those gaps that are smaller than 'maxgap'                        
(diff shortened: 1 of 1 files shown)


pyglider/seaexplorer.py
```diff
@@ -387,6 +387,10 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
                         val = np.datetime64("1970-01-01") + val_us
                     else:
                         val = np.interp(time_timebase.astype(float), time_var.astype(float), var_non_nan)
+
+                        # interpolate only over those gaps that are smaller than 'maxgap'                        
+                        tg_ind = utils.find_gaps(time_var.astype(float),time_timebase.astype(float),maxgap)
```
callumrollo, line 392: I believe this should be de-intented by one level so that it also applies the max gap check to variables with the datetime dtype

callumrollo (changes requested):

pyglider/seaexplorer.py
hvdosser: Good call. Done.

callumrollo (approved):

PR #151, opened 2023-06-07
fix breaking changes in new polars and pin to polars>0.17

callumrollo:
Polars has evolved substantially in the last 6 months with a number of breaking changes

https://github.com/pola-rs/polars/releases

This PR addresses those breaking changes and fixes the environment so that the newer versions of polars are used

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -12,6 +12,6 @@ dependencies:
   - scipy
   - bitstring
   - pooch
-  - polars > 0.17
+  - polars>=0.17
   - pip:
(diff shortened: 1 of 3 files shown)


jklymak:
Hi @callumrollo 

The conda install is failing here.  Are you on a bleeding edge polars?  Perhaps it makes sense to just pin polars to the old version for our use case here?  I guess that makes things more inconvenient if you are using polars as part of your data pipeline, but my understanding was that it is just for speed reasons here.

PR #152, opened 2023-06-13
Added a FillValue for u, v to allow profile submission to IOOS

hvdosser:
As part of the `if` loop where the profile-averaged glider speed _u_ and _v_ are defined, I added an `else` statement to set _u_ and _v_ to a `FillValue` defined in the metadata yaml file, if ocean current speed isn't available (as is usually the case for realtime data).

### DIFF
--- a/pyglider/ncprocess.py
+++ b/pyglider/ncprocess.py
@@ -67,7 +67,7 @@ def extract_timeseries_profiles(inname, outdir, deploymentyaml):
 
                     dss['v'] = dss.water_velocity_northward.mean()
                     dss['v'].attrs = profile_meta['v']
-                else:
+                elif ('water_velocity_eastward' not in dss.keys()) and ('_FillValue' in profile_meta['u']):
                     dss['u'] = profile_meta['u']['_FillValue']
                     dss['u'].attrs = profile_meta['u']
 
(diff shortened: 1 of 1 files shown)


pyglider/ncprocess.py
```diff
@@ -67,6 +67,12 @@ def extract_timeseries_profiles(inname, outdir, deploymentyaml):
 
                     dss['v'] = dss.water_velocity_northward.mean()
                     dss['v'].attrs = profile_meta['v']
+                elif ('water_velocity_eastward' not in dss.keys()) and ('_FillValue' in profile_meta['u']):
```
jklymak, line 70: What is `'_FillValue'` usually?  I'm not familiar with how that gets set up...

jklymak: If this is _not_ meta data do we want to fill with NaN anyways, or just not have the variable at all?

hvdosser: IOOS suggests -999 for a _FillValue, and it is specified in the variable-specific metadata for u and for v. 

I think if we don't specify it in the metadata, we probably don't want the variable at all. That allows people who don't care about e.g. submitting to IOOS to run the code without unused variables taking up space in their netcdf files.

jklymak: Thats fine.  Seems a little strange for FillValue to be a variable-specific metadata, but maybe different sensors need different fill values.

PR #158, opened 2023-06-16
Correct a bug in ncprocess for u,v metadata check

hvdosser:
Fixed the `elif` statement that checks if there is a fill value for _u_ and _v_ in the metadata to prevent a key error if _u_ and _v_ aren't listed as variables in the metadata at all.

### DIFF
--- a/pyglider/ncprocess.py
+++ b/pyglider/ncprocess.py
@@ -71,7 +71,7 @@ def extract_timeseries_profiles(inname, outdir, deploymentyaml):
                     dss['u'] = profile_meta['u'].get('_FillValue', np.NaN)
                     dss['u'].attrs = profile_meta['u']
 
-                    dss['u'] = profile_meta['v'].get('_FillValue', np.NaN)
+                    dss['v'] = profile_meta['v'].get('_FillValue', np.NaN)
                     dss['v'].attrs = profile_meta['v']
 
                 dss['profile_id'] = np.array(p*1.0)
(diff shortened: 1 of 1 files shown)


pyglider/ncprocess.py
```diff
@@ -67,7 +67,7 @@ def extract_timeseries_profiles(inname, outdir, deploymentyaml):
 
                     dss['v'] = dss.water_velocity_northward.mean()
                     dss['v'].attrs = profile_meta['v']
-                elif ('water_velocity_eastward' not in dss.keys()) and ('_FillValue' in profile_meta['u']):
+                elif ('water_velocity_eastward' not in dss.keys()) and (profile_meta.get('u')) and ('_FillValue' in profile_meta['u']):
                     dss['u'] = profile_meta['u']['_FillValue']
                     dss['u'].attrs = profile_meta['u']
```
jklymak, line 72: What you have probably works, but this is more straight forward I think:   Note we know 'water_velocity_eastward' is not in `dss.keys()`

Also maybe we want to error if Fill Value is not there, perhaps with a better error message?  Or maybe you want a different default fill value (-999)

```suggestion
                elif 'u' in profile_meta:
                    dss['u'] = profile_meta['u'].get('_FillValue', np.NaN)
                    dss['u'].attrs = profile_meta['u']
```

hvdosser: Agreed about dropping `water_velocity_eastward`. I like your suggestion to use `np.NaN` rather than an error message.

jklymak (approved):

jklymak:
Feel free to self merge when tests done

PR #173, opened 2024-07-11
Fix tests to passing

cassandra-elmer:
Edits to processing scripts, example NetCDFs, and deployment YAMLs to make tests pass.

The majority of changes are adjusting variable attributes from strings to floats to match existing pyglider code (and the [IOOS GliderDAC requirements](https://ioos.github.io/glider-dac/ngdac-netcdf-file-format-version-2.html)). Other changes include ensuring variables of type Path are input to functions as type str and updating keyword arguments to reflect expected inputs according to current function documentation.


tests/test_seaexplorer.py
```diff
@@ -91,35 +91,35 @@ def test_raw_to_timeseries():
...
+                                                    str(example_dir / 'example-seaexplorer/deploymentRealtime.yml'),
                                                     kind='sub')
     assert 'No such file or directory' in str(missing_file_exc)
     assert result_sub == 'tests/data/l0-profiles/dfo-eva035-20190718.nc'
 
 
 def test_missing_bad_timebase():
     # Prepare yaml files with bad timebase and no timebase
-    with open(example_dir / 'example-seaexplorer/deploymentRealtime.yml') as fin:
+    with open(str(example_dir / 'example-seaexplorer/deploymentRealtime.yml')) as fin:
```
jklymak, line 106: Can you explain why this is needed?  `open` should work with Path variables.

jklymak (commented):
@cassandra-elmer do you mean getting the files to pass the IOOS compliance checker?

I could have sworn the polar changes were already made, but I must not have pushed them...

I'll ope an issue to add the compliance checker to the tests.  I think that can be done routinely so we don't get out of sync with them.

tests/test_seaexplorer.py
cassandra-elmer: > @cassandra-elmer do you mean getting the files to pass the IOOS compliance checker?



The tests I meant are these

Screenshot 2024-07-12 132136

Working on the compliance checker is my own next step, but I wanted to fix the above tests first.



> Can you explain why this is needed? `open` should work with Path variables.



Those are likely redundant because I got trigger happy, yes. There were previous `example_dir / '*.yml` spots that weren't `open` and were breaking and I just carried it through.

tests/test_seaexplorer.py
```diff
@@ -58,12 +58,12 @@ def test_merge_rawnc():
     result_default = seaexplorer.merge_parquet(
             'tests/data/realtime_rawnc/',
             'tests/data/realtime_rawnc/',
-            example_dir / 'example-seaexplorer/deploymentRealtime.yml')
+            str(example_dir / 'example-seaexplorer/deploymentRealtime.yml'))
```
jklymak, line 61: This file still has the `str` conversions as well....  We shouldn't need these.  Thanks!

jklymak (approved):

jklymak:
Thanks @cassandra-elmer !

PR #183, opened 2024-08-06
Support python 3.11 & 3.12

callumrollo:
This PR expands the test grid to support Python 3.11 and 3.12

It also pins the recent stable 1.x version of polars, rather than the dev 0.x version. 

It pins numpy < 2 as the the numpy 2 release, among other things, deprecates `np.NaN` for `np.nan`, which will necessitate some refactoring https://numpy.org/devdocs/release/2.0.0-notes.html

This PR disables the cache so that new versions of polars etc. are fetched. An alternative is to move to micromamba for the CI, which should give a speed improvement over conda

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v2
       - name: Cache conda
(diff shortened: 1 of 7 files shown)


jklymak:
So I thought all the changes for numpy 2.0 were in already.  Of not maybe just fix them?  

I'm pretty surprised by the 2.0 change as I don't remember np.NaN being deprecated and I don't understand why it needed to be removed.

callumrollo:
Yeah I'm not sure why they felt the need to remove it, but it's an easy backward compatible change for us to make replacing np.NaN with np.nan. I'll replace it in this PR and unpin numpy in the requirements

jklymak:
I must have forgotten to push some of these changes as I'd already been using 2.0.   But that's fine I can rebase

jklymak (approved):

PR #184, opened 2024-08-06
move tests to micromamba and update gh-actions versions

callumrollo:
The tests here will fail as they default to numpy 2. Need to either refactor np.NaN >> np.nan or merge #183 first


callumrollo:
I think codecov has been silently broken for a while. This PR would bring it back into service

jklymak (approved):

jklymak:
Thanks @callumrollo

PR #209, opened 2024-12-12
Migrate to `pyproject.toml`, update min Python to 3.9, consolidate dev workflows

VeckoTheGecko:
Changes:
- Migrate to `pyproject.toml`
- Update min Python version to 3.9
- Add pypi metadata classifiers
- Add contributing guide
- Consolidate installation (for users, testing and docs)
- Dependabot config
- Fix `conf.py`
- Clean up `tests.yml`
- Remove circleci
- Add `tests.utils` to consolidate test path parameters

Can we add [pre-commit CI](https://pre-commit.ci/)* to this project, as well as set up [docs building for PRs via RTD](https://docs.readthedocs.io/en/stable/guides/pull-requests.html)**?

\* In preparation for when we add pre-commit config in next PR  
\*\* I think this is a better way to check that the docs build. It allows previewing the docs as well (which I don't think the circleci build did).


jklymak:
So what is the plan here?  Should we just merge this and see if it works, and then fix any issues?

jklymak:
Only question I had was the empty `__init__.py` file in `tests/`?

jklymak:
... also - what is the plan for changing the files to ruff?  Is that a second PR?  I guess I thought that would come _before_ this PR, not after.

VeckoTheGecko:
> So what is the plan here? Should we just merge this and see if it works, and then fix any issues?

Sounds good. Any particular issues that you're concerned about?

> Only question I had was the empty `__init__.py` file in `tests/`?

This makes `tests` a package, allowing us to do `from tests.utils import ...` (consolidating `LIBRARY_DIR` to be stored in one place).

> ... also - what is the plan for changing the files to ruff? Is that a second PR? I guess I thought that would come _before_ this PR, not after.

I think this PR first and then #210 would be good. Once this is merged, #210 should only be 2 commits so shouldn't be a need to rebase

PR #210, opened 2024-12-12
Ruff config, pytest config, pre-commit config

VeckoTheGecko:
Blocked by #209. Once its merged I'll rebase this onto it and run the tooling. 64ec4af and after are the commits that are part of this PR

Added:
- Ruff config
- pre-commit config
- pytest config

PR #212, opened 2024-12-12
add Ocean Gliders format output option to seaxexplorer processing

### LINKED ISSUE #21: Output OG 1.0
Ultimately, we should aim for the OceanGliders 1.0 format https://github.com/OceanGlidersCommunity/OG-format-user-manual

This will take some time. If you spot an atomic change that can be made to move toward OG 1.0, please submit an Issue for it

callumrollo:
This PR is under development and is waiting on several currently active proposed edits to the Ocean Gliders format before it can be finalised 

This will resolve #21

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -304,7 +304,7 @@ def _remove_fill_values(df, fill_value=9999):
 
 def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
                       profile_filt_time=100, profile_min_time=300,
-                      maxgap=10, interpolate=False, fnamesuffix=''):
+                      maxgap=10, interpolate=False, fnamesuffix='', og_format=False):
     """
     A little different than above, for the 4-file version of the data set.
     """
@@ -320,7 +320,12 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
     _log.info(f'Opening combined payload file {indir}/{id}-{kind}pld.parquet')
     sensor = pl.read_parquet(f'{indir}/{id}-{kind}pld.parquet')
     sensor = _remove_fill_values(sensor)
-
+    if og_format:
+        # temporarily translate from og names to pyglider names
+        for pyglider_var, og_var in utils.pyglider_og_var_dict.items():
+            if og_var in ncvar.keys():
+                ncvar[pyglider_var] = ncvar.pop(og_var)
+        
     # build a new data set based on info in `deploymentyaml.`
     # We will use ctd as the interpolant
     ds = xr.Dataset()
@@ -470,8 +475,6 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
     ds = ds.assign_coords(longitude=ds.longitude)
     ds = ds.assign_coords(latitude=ds.latitude)
     ds = ds.assign_coords(depth=ds.depth)
-    # ds = ds._get_distance_over_ground(ds)
-
     ds = utils.fill_metadata(ds, deployment['metadata'], device_data)
 
     start = ds['time'].values[0]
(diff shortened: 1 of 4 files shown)


jklymak:
A quick skim - I'm not clear that a lot of this cannot be handled via the yaml.  

For the rest, it would be nice to keep new variables like platform id in the yaml as well. I've not looked but perhaps that requires a new variable type that just accepts a constant value.  

Maybe we should have a quick call, but I would rather see a general approach rather than special "og" carve outs

pyglider/seaexplorer.py
```diff
@@ -485,18 +488,26 @@ def raw_to_timeseries(indir, outdir, deploymentyaml, kind='raw',
...
+        outname = f"{outdir}/{ds.attrs['id']}.nc"
+        _log.info('writing %s', outname)
+        ds.to_netcdf(outname, 'w',
+                     encoding={'TIME': {'units': 'seconds since 1970-01-01T00:00:00Z',
+                                        'dtype': 'float64'}})
+    else:
+        outname = outdir + id0 + fnamesuffix + '.nc'
+        _log.info('writing %s', outname)
+        ds.to_netcdf(outname, 'w',
+                     encoding={'time': {'units': 'seconds since 1970-01-01T00:00:00Z',
```
jklymak, line 509: For example, 'time'/'TIME' could be replaced by a time-encoder yaml entry. eg calls to `ds.time` could be replaced by `ds[time_var_name]` or similar?  

Not sure about 'latitude' etc, but why do they need to be translated back and forth?  For distance over ground?  Maybe just use inside that function?

pyglider/utils.py
```diff
@@ -633,6 +633,87 @@ def find_gaps(sample_time, timebase, maxgap):
...
+        attrs={
+            "long_name": "model of the glider",
+            "platform_model_vocabulary": "None",
+        },
+    )
+    ds["PLATFORM_SERIAL_NUMBER"] = xr.DataArray(
+        f"sea{ds.attrs['glider_serial'].zfill(3)}",
+        attrs={"long_name": "glider serial number"},
+    )
+    ds["DEPLOYMENT_TIME"] = np.nanmin(ds.TIME.values)
```
jklymak, line 696: The yamls have a "conversion" field for each variable that tries to fine the method in `utils.py`.  So this _could_ look like 

```
netcdf_variables:
    DEPLOYMENT_TIME:
         "long_name": "date of deployment"
         ...
         "conversion": "nanmin_time"
```

with the accompanying change in `utils.py`.  We may want to think of a way to make that more general `(np.nan_min, "DEPLOYMENT_TIME)` if yaml accepts tuples.

VeckoTheGecko:
66519318 (via #210) introduced a bunch of formatting changes that will likely cause merge conflicts. Here are merge and rebase workflows to avoid these conflicts (it depends on having `pre-commit` installed - see the contributing docs).

I recommend first creating a `temp` branch off of your branch of interest to test these workflows and make sure the output is as expected.

```bash
# merge all the way up to the commit that introduced the formatting changes
git merge 66519318^1

# merge the commit that introduced the formatting changes, prioritising the working branch during conflicts
git merge -s ours 66519318 --no-commit 
pre-commit run --all-files
git commit --no-edit

# merge the rest of the way
git merge main
```

or by doing a rebase (if you want a clean history)

```bash
# Rebase up to the commit introducing formatting changes
git rebase 66519318^1

# Rebase onto the commit introducing formatting changes.
# Accept working branch changes (i.e., "theirs" during rebase
# - different to merge. See git rebase docs) during conflicts.
# During the rebase we will run the formatter after applying
# changes from commits, before committing so that everything is compliant.
git rebase -X theirs -i 66519318
# Replace all "pick" with "edit" or "e". Start the rebase...

# Repeatedly run
git add -u && pre-commit run --all-files && git rebase --continue
# to run pre-commit tooling, stage modified files, and continue
# until rebase is complete. Sometimes pre-commit can't autofix
# issues, then you need to fix them manually.
# If you make a mistake you can always do `git rebase --abort`

# Finally, rebase onto main
git rebase main
```

pyglider/seaexplorer.py
callumrollo: I can localise lon/lat to just change within those functions, I'll make a new PR with `ds[time_var_name]` and see what it looks like

PR #213, opened 2024-12-18
Fix bug: set `have_dbdreader` to `False` if it's not installed

santisoler:
Set `have_dbdreader` to `False` when it cannot be imported, so `binary_to_timeseries()` raises the proper `ImportError` instead of error out because the `dbdreader` variable is not defined.

### DIFF
--- a/pyglider/slocum.py
+++ b/pyglider/slocum.py
@@ -12,7 +12,7 @@ try:
 
     have_dbdreader = True
 except ImportError:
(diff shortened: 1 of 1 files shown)


jklymak:
Thanks that was indeed a typo.  OTOH, I'm not sure if we should continue to support dbd reading without dbd reader.

santisoler:
No worries! Glad to contribute 🛩️

PR #215, opened 2025-01-08
sea explorer white space in payload files

clayton33:
We have a number of missions from year 2018 with navigation firmware 2.1.5 and payload firmware 2.1.1. In the raw payload files there are leading whitespaces in both the data and the header which resulted in an error being thrown when converting the data to Float64. This fix resolves that issue.

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -158,6 +158,12 @@ def raw_to_rawnc(
                             )
                         )
                         out = out.rename({'PLD_REALTIMECLOCK': 'time'})
+                    # remove leading and trailing whitespace
+                    # from entire dataframe
+                    for col_name in out.columns:
+                        out = out.with_columns(pl.col(col_name).str.strip_chars())
+                    # from column names
+                    out = out.rename(str.strip)
                     for col_name in out.columns:
                         if 'time' not in col_name.lower():
                             out = out.with_columns(pl.col(col_name).cast(pl.Float64))
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -143,7 +143,6 @@ def raw_to_rawnc(
                         _log.warning(f'Could not read {f}')
                         badfiles.append(f)
                         continue
-
                     # Parse the datetime from nav files (called Timestamp) and pld1 files (called PLD_REALTIMECLOCK)
                     if 'Timestamp' in out.columns:
                         out = out.with_columns(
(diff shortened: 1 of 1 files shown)


jklymak:
Can you explain the data that causes this error a little more explicitly.  The added code looks potentially slow and I wonder if there is a better way to strip leading white space available in polars.

clayton33:
[exampleData.zip](https://github.com/user-attachments/files/18350266/exampleData.zip)
Attached is some example data.

jklymak:
Thanks, but what about this data is causing the error?  Can you just extract the string that is causing the problem so we can look for solutions?

clayton33:
There are leading whitespaces throughout the entire file and it's in every payload file, so I think looking at every column that is a `str` type makes sense. Any files after that navigation and payload firmware don't have that issue, and they'll read in as `float64`, so these newly added lines will ultimately do nothing in those cases since there won't be any columns of `str` type.

jklymak:
Can you include here a string that has a leading white space?  I opened a couple of those files and I'm not seeing any leading whitespaces, but maybe we are talking about different things.

jklymak:
Also, maybe this is a windows/Unix thing?  I'm on a Mac and don't see any leading spaces.  Perhaps we need to specify the encoding of the files when we open them.  It's super unlikely that polars can't handle white space in float conversion, so I think it must be something like that.

clayton33:
Sorry for the confusion. I sent data for the wrong mission (we had mission 32 in the same year for two different gliders).

[exampleData.zip](https://github.com/user-attachments/files/18351350/exampleData.zip)

richardsc:
@clayton33 can you also post the error that you get when running with our yaml?

clayton33:
This is the error that I get

```
polars.exceptions.InvalidOperationError: conversion from `str` to `f64` failed in column ' NAV_RESOURCE' for 58 out of 58 values: [" 105", " 105", … " 117"]
```

jklymak:
https://github.com/pola-rs/polars/issues/10587 maybe

clayton33:
Yes, I had come across that issue. I followed the suggestion from this comment https://github.com/pola-rs/polars/issues/10587#issuecomment-2116459477

clayton33:
Checking in to see if there has been any thought into incorporating this change.

jklymak:
Its fine by me, but would be good if @callumrollo chimed in

richardsc:
Hi @jklymak, @callumrollo! Any thoughts on merging this PR? We've been using it in production and it all seems good on our files (with and without white space).

richardsc:
Thanks!

jklymak:
Yes sorry to be slow.  Always feel free to ping!

PR #216, opened 2025-02-26
Fix deployment_start and _end to be different and not clipped

cassandra-elmer:
Both start and end times in the timeseries NetCDF metadata were the same and missing the last digit.


cassandra-elmer:
Should be all good now! Wasn't sure if regenerating was something you wanted to do or if I should do.

jklymak:
Thanks a lot!

Issue #221, opened 2025-06-04
Small changes - check for dataset vars, and metadata


smwoodman:
happy to split these out into their own issues

The functionality in https://github.com/c-proof/pyglider/pull/217, of checking for vars before running `utils.get_glider_depth` and `utils.get_derived_eos_raw`, is nice as it simplifies making an engineering timeseries by not requiring these variables to be in and engineering yaml. However, there are still a few `_log.debug` lines that expect eg `ds.pressure`, and thus throw an error if that variable is not in the yaml file and thus dataset.

Also, around `utils.fill_metadata`:
- the code `ds.attrs['deployment_name'] = metadata['deployment_name']` duplicates [this line](https://github.com/c-proof/pyglider/blob/90e7c0461ac75692f27fe72b01eeed203337d6a7/pyglider/utils.py#L542)
- [this line](https://github.com/c-proof/pyglider/blob/90e7c0461ac75692f27fe72b01eeed203337d6a7/pyglider/utils.py#L547C48-L547C57) `ds.attrs['standard_name_vocabulary'] = 'CF STandard Name Table v72'` has a typo
- it would be nice in `ds.attrs['history'] = 'CPROOF glider toolbox version: pre-tag'` ([here](https://github.com/c-proof/pyglider/blob/90e7c0461ac75692f27fe72b01eeed203337d6a7/pyglider/utils.py#L541)) to get the actual pyglider relase
- Can/should `ds.attrs['deployment_start'] = str(start)[:19]` and `ds.attrs['deployment_end'] = str(end)[:19]` move from `slocum.binary_to_timeseries` to `utils.fill_metadata`, for consistency?

smwoodman:
Closed in #222

PR #222, opened 2025-06-05
small changes for metadata and debug calls

smwoodman:
pr for #221 

cleanup of metadata: typo fix, include pyglider version

cleanup of debug calls in binary_to_timeseries, so that deploymentyaml doesn't have to have pressure for get_glider_depth, or temp/conductivity/pressure for get_derived_eos_raw

### DIFF
--- a/pyglider/slocum.py
+++ b/pyglider/slocum.py
@@ -971,11 +971,13 @@ def binary_to_timeseries(
         attrs = utils.fill_required_attrs(attrs)
         ds[name] = (('time'), val, attrs)
 
-    _log.info(f'Getting glider depths, {ds}')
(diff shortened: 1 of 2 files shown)


jklymak:
Yeah the tests are a little suboptimal in that if you fix anything you need to copy new files to `tests/expected`.  Probably should write a helper function to do that.

smwoodman:
Ok - I've updated the relevant expected files, and the tests past per [here](https://github.com/smwoodman/pyglider/actions/runs/15480001884).

jklymak (approved):

jklymak:
Thanks @smwoodman

Issue #220, opened 2025-06-04
gridded data - incorrect profile_time_start and profile_time_end


smwoodman:
In `make_gridfiles` output, the profile_time_start and profile_time_end values are the same values as latitude. It looks like the min/max times for each profile still need to be separately calculated

jklymak:
I'm not clear what you mean here?

smwoodman:
In the NetCDF file created by `make_gridfiles`, the values in the profile_time_start and profile_time_end data variables are the same values as those in the latitude variable. This is because in this chunk

``` python
    for td in ('time_1970', 'longitude', 'latitude'):
        good = np.where(~np.isnan(ds[td]) & (ds['profile_index'] % 1 == 0))[0]
        dat, xedges, binnumber = stats.binned_statistic(
            ds['profile_index'].values[good],
            ds[td].values[good],
            statistic='mean',
            bins=[profile_bins],
        )
        if td == 'time_1970':
            td = 'time'
            dat = dat.astype('timedelta64[ns]') + np.datetime64('1970-01-01T00:00:00')
        _log.info(f'{td} {len(dat)}')
        dsout[td] = (('time'), dat, ds[td].attrs)
    ds.drop('time_1970')
    good = np.where(~np.isnan(ds['time']) & (ds['profile_index'] % 1 == 0))[0]
    _log.info(f'Done times! {len(dat)}')
    dsout['profile_time_start'] = ((xdimname), dat, profile_meta['profile_time_start'])
    dsout['profile_time_end'] = ((xdimname), dat, profile_meta['profile_time_end'])
```

the `dat` variable still has the latitude values when it is used for both profile_time_start and profile_time_end.

I think it will be simple to add profile_time_start and profile_time_end to the for loop, along with a bit of code inside to calcualte the time mins/maxes? I'll draft something soon.

Issue #219, opened 2025-06-04
gridded data - user-provided depth bins


smwoodman:
Hi @jklymak 
We're close in our NOAA group to having a standard workflow using pyglider + a [toolbox](https://github.com/SWFSC/glider-utils) with wrapper/post-processing for processing our slocum deployments. We're really appreciating all the pyglider functionality. I'm going to be submitting a few issues with requests for generalizing some of the pyglider functions, for which I'm happy to submit PRs. 

In `ncprocess.make_gridfiles`, currently the user can provide the depth bin size, but the bins themselves are hardcoded to be between 0.5 and 1100m. We'd like to be able to both start at 0m (for compatibility with some past processing), and extend the bins to 1200m to be able to capture all info from downward-facing acoustic instruments on the glider.

Would you be good with adding a `depth_bins` argument to `make_gridfiles`? I'm imagining a `depth_bins=None` being the default, in which case the current behavior happens. Or the user could provide their own array.

jklymak:
Yes that seems reasonable.  Perhaps even deprecate dz and let `depth_bins = dz` be a shorthand. 

Note the bins actually go from 0-1 m, 1-2 m etc, and its th bin centres that are 0.5 etc...

jklymak:
Thats almost certainly a bug.   

should be:

```
    depth_bins = np.arange(0, 1100.1, dz)
    depths = depth_bins[:-1] + 0.5*dz
```

smwoodman:
Cool, thanks. I'll hopefully get to this in the next few days

PR #223, opened 2025-06-06
gridding: allow user depth bins and fix profile times

smwoodman:
A first pass at #219 and #220 

In the pr, I left the depth values as the start of the bin for user-provided bins. This is more intuitive to me and seems more widely used in my experience, but makes it inconsistent with both past behavior and behavior using dz. Let me know your thoughts?


pyglider/ncprocess.py
```diff
@@ -266,14 +288,28 @@ def make_gridfiles(
...
         dsout[td] = (('time'), dat, ds[td].attrs)
-    ds.drop('time_1970')
+
+    # Profile start and end (min/max) times
+    profile_lookup = {'profile_time_start': "min", 'profile_time_end': "max"}
     good = np.where(~np.isnan(ds['time']) & (ds['profile_index'] % 1 == 0))[0]
-    _log.info(f'Done times! {len(dat)}')
-    dsout['profile_time_start'] = ((xdimname), dat, profile_meta['profile_time_start'])
-    dsout['profile_time_end'] = ((xdimname), dat, profile_meta['profile_time_end'])
+    for td, bin_stat in profile_lookup.items():
```
jklymak, line 295: Can you walk me through what happened before, and what happens now here?

smwoodman: For sure. Before, at that point in the function `dat` was still the value from the `for td in ('time_1970', 'longitude', 'latitude'):` loop. So, in practice these lines were assigning the mean latitude values, binned by profile index, to profile_time_start and profile_time_end. 

The new chunk instead calculates the min (for profile_time_start) and max (for profile_time_end) times for each profile bin. Since these are now times rather than floats, they also need to be encoded when writing the netcdf

smwoodman:
updates made

jklymak (approved):

Issue #226, opened 2025-06-21
order of dimensions in gridded data output


smwoodman:
A colleague pointed out that the gridded NetCDF file produced by `make_gridfiles` has depth as the 'x coordinate', and time as the 'y coordinate'. 

https://github.com/c-proof/pyglider/blob/c1218acb321f7b2ba9533de8cd6505bddfaeca7e/pyglider/ncprocess.py#L273-L276

It feels more intuitive to have these switched? For instance, in the function itself time is referred to as the `xdimname`. 
Also, this (time as the x dim, depth as the y dim) is how the old [socib toolbox](https://github.com/socib/glider_toolbox) writes its gridded data files.

I believe the only change would be changing line 275 to `coords={'profile': (xdimname, profiles), 'depth': ('depth', depths)}`

jklymak:
I'd not be in favour of this change. This is just a convention and in my mind Depth are rows, and profiles are columns because if you look at it as a spreadsheet, data going down a column is increasing in depth.

smwoodman:
I had mostly thought about it from a plotting perspective, where (afaik) time or profile is always on the x-axis, and depth is on the y-axis. But fair enough, and thanks for considering.

jklymak:
> I had mostly thought about it from a plotting perspective, where (afaik) time or profile is always on the x-axis, and depth is on the y-axis. But fair enough, and thanks for considering.

Yeah, except I think you have this backwards.  At least Matlab and Matplotlib will plot the rows as y-axis and the columns along the xaxis.  Eg `plt.pcolormesh(ds.temperature.values)` looks correct (except it is upside down).  


> Trying to catch or classify all variables (eg, waypoint_longitude and waypoint_latitude) that a user could make part of the timeseries dataset seems like too much. However, it does feel like make_gridfiles should handle profile_index and profile_direction, especially since profile_index is required by the function. For instance, there could be a if k in ['profile_index', 'profile_direction']: ... right after line 324. Or, they could be worked into the profile_lookup section beginning in [line 305](https://github.com/c-proof/pyglider/blob/c1218acb321f7b2ba9533de8cd6505bddfaeca7e/pyglider/ncprocess.py#L305C5-L305C20).

I think its makes sense that the `profile_x` variables should be part of the default exclusion.

I think we could easily also make the exclusion list a kwarg or info that gets specified in the yaml somehow.  I'm not sure of the details, but I think we could easily accommodate this.

PR #227, opened 2025-07-09
gridded: dimensions and variable exclusion

smwoodman:
Addresses #226


smwoodman:
@jklymak a couple fyis to explain my thinking. Happy to adjust these as you like. Sorry, not sure how to directly embed lines from the pr in the comment

> `ds['time_1970'] = ds.longitude.copy()`

Changing this from `ds.temperature` removes temperature as a 'required' variable for this function to run

> profile_direction

Profile direction felt useful to add, as an indicator of if a profile was while the glider was ascending or descending. Since it was only binned by profile, it seemed easiest to put direction with the time/lat/lon binning

> exclude_vars , `exclude_vars += list(dsout.keys()) + ["depth"]`

It felt simplest to let the user pass a list of variables (exclude_var) to ignore when gridding, rather than eg looking for a new 'grid_ignore: true' key in the deploymentyaml file. This line, along with the check `if (k in exclude_vars) or ('time' in k) or ('profile' in k): ... continue` should be consistent with the current `make_gridfiles` behavior. It also ensures that both a) all vars already in dsout, like lat/lon or whichever else are added in the future, and b) all time/profile variables, aren't gridded

> try: dsout['waypoint_latitude'].attrs.pop('standard_name') ...

`profile_time_start` and `profile_time_end` are guaranteed to be in dsout, since they're made in this function. Thus, I moved them to the top of the try statement, in case waypoint_latitude and waypoint_longitude aren't in dsout

smwoodman:
> One other question. Is there a reason that `dsout['profile'].attrs['cf_role'] = 'profile_id'` is the only profile attribute? Could there be a line such as `dsout['profile'].attrs = ds.profile_index.attrs` to bring along that info?

@jklymak sorry for the ping, but I wanted to check in about both this question, and the pr as a whole?

pyglider/ncprocess.py
```diff
@@ -319,8 +338,12 @@ def make_gridfiles(
...
     _log.info(f'Done times!')
 
-    for k in ds.keys():
-        if k in ['time', 'profile', 'longitude', 'latitude', 'depth'] or 'time' in k:
+    if exclude_vars is None:
+        exclude_vars = []
+    exclude_vars += list(dsout.keys()) + ["depth"]
+    for k in ds.keys():       
+        if (k in exclude_vars) or ('time' in k) or ('profile' in k):
+            _log.debug('Not gridding %s', k)
```
jklymak, line 346: This looks to be gridding longitude and latitude by deafult now?  Or are they in the default `exclude_vars`?  

I wonder if for non-coordinate data (eg temperature etc) we should really be setting whether to grid it or not in the yaml instead of specifying a list here.  eg in the yaml say `grid_exclude: True` if we don't want it to be gridded, but we do want it in the time series?

smwoodman: > This looks to be gridding longitude and latitude by deafult now? Or are they in the default `exclude_vars`?

longitude and latitude (along with time, profile_direction, profile_time_start, and profile_time_end) are all guaranteed to be part of dsout. Thus, they're always added to `exclude_vars` in `exclude_vars += list(dsout.keys()) + ["depth"]`

smwoodman: > I wonder if for non-coordinate data (eg temperature etc) we should really be setting whether to grid it or not in the yaml instead of specifying a list here. eg in the yaml say `grid_exclude: True` if we don't want it to be gridded, but we do want it in the time series?

A `grid_exclude` boolean key feels totally reasonable to me. It's also more consistent with the function checking for the 'average_method' key. I'll update the pr now

smwoodman:
@jklymak wanted to check in about this, as I'm currently working on some code that uses these features. 

But, please let me know if you're busy at the moment, and prefer that either I ping someone else or we pick this back up at a later date!

pyglider/ncprocess.py
jklymak: @smwoodman  I was waiting to understand this...

jklymak:
@swoodman, sorry this has dragged on...  Is this ready to go from your point of view?  I think I'm fine with these changes.

smwoodman:
@jklymak hey, thanks, and no worries at all. I was away for fieldwork for most of the winter, and so have only been getting back into glider world recently. 

I haven't done a deep dive into the changes since being back. How about I do that, put a brief summary here for you can sanity-check, and we can go from there? I see there's at least one other open gridding pr, so I'll make sure to get to this in the next few days

smwoodman:
Hey @jklymak , sorry this took me so long to get to. I dove back in today, and feel good about the changes. Figured I'd start with all notes in one place; happy to flag any specific spots in the changed file though

Summary of changes:

-	Added docs 
-	Remove dependence on `ds.temperature`; instead do `ds['time_1970'] = ds.longitude.copy()`, since longitude is implicitly required elsewhere
-	profile_direction now only has one dimension (time)
-	Use `ds.drop_vars` instead of `ds.drop`, to fix xarray warning
-	Use dictionary and tuples to have a single loop to process all vars that should only have time as coordinate
-	Always exclude 'distance_over_ground' (because not appropriate) and 'profile_index' (to not have duplicate with 'profile') from gridded output
-	Slight reorganizing to put all `dsout['profile_time_start']` cleanup in the same try/except block
-	For each variable, always describe gridding method in 'processing' attribute. Use += if 'processing' attribute already exists, so as to not overwrite existing values

Questions:

- Right now, the function checks the timeseries dataset for a user-specified ‘average_method’ (`if 'average_method' in ds[k].attrs`), and the deploymentyaml for grid_exclude (`if k in ncvar and 'grid_exclude' in ncvar[k]...`). It feels like these should be consistent, ie both check either the dataset or ncvars? What would you prefer?
- Currently 'average_method' info (from the deployment yaml) is written to an attribute named 'processing'. From past work, I know that SOCIB uses the attribute 'grid_method` for this info, which I feel is a bit more explicit than 'processing'. Do you prefer to stick with 'processing'?

smwoodman:
Hey @jklymak, I wanted to check in about this PR?

Looks like you've been doing lots of work in pyglider. If you feel any of the changes here are still useful/relevant, I'm happy to catch this PR up with the current pyglider

jklymak:
Hi @smwoodman , yeah soon would be good.  Let's see if #240 drops and then we can work this in rebased on that.  After that I'll do another release (while I remember how!).  Please ping if you see #240 merged.

jklymak:
OK let's sort this out.  #240 is in.  and the QC changes are in.

The yaml _already_ accepts an `average_method` that is supposed to guide how the gridding is done on a variable.  I wonder if much of this PR could be accomplished by some defaults to that averaging method for the "special" variables and/or letting the user specify the averaging method in the yaml?  Given that #240 is in, I'd prefer we didn't rely on variable names to decide how to do average things.  

Thanks!

smwoodman:
Thanks for checking in, and sorry I'm so late responding here

I don't believe this PR changes the averaging method from being specified in the yaml. My main aims were to 1) provide a way to allow the user to exclude a variable from gridding, eg a 'grid_exclude' flag in the yaml, 2) better handle 'distance_over_ground' and 'profile_direction' if they exist in the input timeseries, and 3) have the average_method always be written out in the variable attributes , including for a simple mean. 

For 2, this means don't grid 'dog', and ensure 'profile_direction' only has a time dimensions
For 3, I guess that could just be up to the user, ie if they want that attribute in the netcdf then they need to include average_method in their original yaml file. 

later this morning I'll working on getting this pr up to date with v0.9 for you to consider, but of course please let me know if any of this doesn't make sense to you, or you have other preferences.

jklymak:
>1) provide a way to allow the user to exclude a variable from gridding, eg a 'grid_exclude' flag in the yaml, 2) better handle 'distance_over_ground' and 'profile_direction' if they exist in the input timeseries, and 3) have the average_method always be written out in the variable attributes , including for a simple mean.

1) what if we set averaging method to "dont-grid" or "null" and that would be a signal to not add to the grid?
3) sounds reasonable to me. 

2) I'm not clear on "better handle" here.  I'd suggest you make a different PR for that so we don't get mixed concerns.  

Thanks and sorry for the process.

smwoodman:
> Thanks and sorry for the process.

Not at all - thank you for all your time here!

1. That sounds good to me. I'll go with `averaging method = 'none'` for now, but let me know if you have another preference

3. Sounds good - I'll close this pr, and start a new one once I manage to update the tests.

PR #240, opened 2026-06-19
ENH: generalize derived quantity and dimension names

jklymak:
## Add OG 1.0 output support

This PR adds support for writing OceanGliders 1.0 format output from both Slocum and SeaExplorer gliders, alongside the existing IOOS GDAC format. The core change is a YAML-driven approach to variable roles and derived-variable computation that replaces hardcoded variable names and fixed processing pipelines.

### New YAML keys

Two new keys in `netcdf_variables` control OG 1.0 behaviour.

**`processing_role`** tells the pipeline what role a variable plays (time, latitude, longitude, pressure, depth, conductivity, ...) so that variable names are resolved symbolically rather than by hardcoded strings:

```yaml
CNDC:
  source:        sci_water_cond
  processing_role: conductivity
TEMP:
  source:        sci_water_temp
  processing_role: temperature
```

**`processing_method`** replaces a direct sensor `source` with a named computation and its named inputs. For example, salinity is declared entirely in the YAML:

```yaml
PSAL:
  processing_method:
    practical_salinity:
      conductivity: CNDC
      temperature:  TEMP
      pressure:     PRES
```

The same pattern drives `potential_temperature`, `potential_density_sigma0`, `density`, and `distance_over_ground`. Each result gets provenance attributes (method, sources, comment) set automatically.

A top-level `output_dimension: N_MEASUREMENTS` key renames the time dimension in the output file to match the OG 1.0 convention. Existing GDAC-style YAMLs without these keys are unchanged.

### Implementation

**`utils.py`** gains four new helpers: `_get_varnames` builds a role→name dict from `processing_role` entries (falling back to variable-name-as-role for GDAC YAMLs); `_load_dataset` / `_save_dataset` normalise `output_dimension` on read/write; `_dispatch_processing_methods` executes `processing_method` entries in YAML order with provenance. Existing utility functions (`get_glider_depth`, `get_distance_over_ground`, `get_derived_eos_raw`, `get_profiles_new`, `fill_metadata`) gain an optional `varnames=` argument.

Both `binary_to_timeseries` (slocum.py) and `raw_to_timeseries` (seaexplorer.py) now resolve variable names via `_get_varnames`, skip `processing_method` variables in the sensor loop and dispatch them afterwards, and use `_save_dataset` for output.

`make_gridfiles` in ncprocess.py uses `_load_dataset` and resolves depth/profile/lat/lon names from roles. `profile_variables` is now optional.

### Examples and tests

- `tests/example-data/example-slocum-og10/` — full Slocum OG 1.0 example with YAML and processing script
- `tests/example-data/example-seaexplorer/deploymentRealtime_og10.yml` — SeaExplorer OG 1.0 example
- New test


jklymak:
@callumrollo @clayton33 @richardsc 

This is my (AI aided) attempt to make the processing pipeline more flexible.  In particular see the new yamls in `tests/example_data/example_seaexplorer/deploymentRealtime_og10.yml` and `tests/example_data/example_seaexplorer/process_deploymentRealtime_og10.py`.  

I doubt that all the data in the yaml is exactly OG1.0 compliant, but that should be easy to fix.  It is more getting rid of the internal dependence on variables like `temperature` that is the goal here.

jklymak:
@callumrollo I couldn't find an OG1.0 compliance checker anywhere.  Is there one?

clayton33:
A compliance checker is linked in the README page here : https://github.com/OceanGlidersCommunity/OG-format-user-manual

jklymak:
Oh it's just the same compliance checker as the ioos one.  Great!

callumrollo:
Yes that's the one! It's a plugin I've been developing with a couple of other contributors. I can't guarantee we're hitting everything, but it covers all the "mandatory" requirements of OG1 compliance

jklymak:
Ok this PR uses that compliance checker. 

I did open an issue on the og 1.0 page but perhaps should be on the plugin page. We have some variables that og1 doesn't recognize but it comes up has high priority errors if the vocabulary doesn't exist.  So I ignore those.  I'm not quite clear what role folks foresee  for vocabularies or why it would be a problem to not have one.

jklymak:
@callumrollo  @clayton33 

Did either of you get a chance to look at this.   In particular, whether the yaml way of driving things makes sense?

callumrollo:
I think the yaml-driven approach is really neat! Keeps with the existing style of pyglider. I'm traveling at the moment so have limited ability to test, but will try running a test dataset through this tomorrow.

callumrollo:
Managed to run it on one of our datasets with the adapted yaml. I think the option to add custom conversion methods is very neat. Currently missing the global attribute `date_created` in the expected file `expected/example-seaexplorer/L0-timeseries-og10/dfo-eva035-20190718.nc` but these probably just need to be re-run as it's present in the output of the example in the tests

Overall looks great!

EllisK-L:
Hey, 
This is great, I've been testing this out for one of our slocum datasets, I also really like this solution. I was able to generate a fairly complete OG1.0 dataset pretty straightforwardly. I'm just wondering about the calculated variable names. Maybe it should set the variable names to the OG names if it detects it's an OG dataset so the user doesn't have to specify/override all the calculated fields in order for it to be OG compliant. For example setting profile_index to PROFILE_NUMBER automatically.
Curious on what others think.

jklymak:
Thanks for testing both!

@EllisK-L can we express that somehow in the yaml?  I think so actually, so let me give that a shot.  Currently we hardcode some things to `ds.profile_index`, but I don't see why we couldn't have either flexibility in that name, or a derived variable that is specified in the yaml and consistent w/ whatever OG1.0 wants.

EllisK-L:
Oh yeah I see that one of the `ds.profile_index` lines is in a function that doesn't have the deployment yaml. Maybe it could have an optional argument for it being an OG dataset, then just go up the stack until the caller has a deployment yaml so it can check if it's an OG dataset or not. 
Or it could check if the column "profile_index" or "PROFILE_NUMBER" exists in the dataset, and use whichever it finds.

jklymak:
@EllisK-L 

Actually, I looked again at what this does - it already produces `PROFILE_DIRECTION` and `PROFILE_NUMBER` compliant with OG1.0.  Note you need 
```
  # -------------------------------------------------------------------------
  # Derived coordinate variables — processing_method specifies inputs
  # -------------------------------------------------------------------------

  DEPTH:
    processing_method:
      depth_from_pressure:
        pressure: PRES
        latitude: LATITUDE
    processing_role: depth
    long_name:     Depth below surface of the water body
    standard_name: depth
    units:         m
    axis:          Z
    positive:      down
    reference_datum: sea-surface
    observation_type: calculated
    valid_min:     0.0
    valid_max:     2000.0
    vocabulary:    http://vocab.nerc.ac.uk/collection/OG1/current/DEPTH/

  PROFILE_NUMBER:
    processing_method:
      find_profiles:
        pressure: PRES
    processing_role: profile_index
    long_name:     Profile number
    comment:       "PROFILE_NUMBER increments by one each time the glider starts
                    an ascending or descending profile."
    valid_min:     1
    valid_max:     2147483647

  PROFILE_DIRECTION:
    processing_method:
      find_profiles:
        pressure: PRES
    processing_role: profile_direction
    long_name:     Vertical direction of profile
    comment:       "1 = descending, -1 = ascending, 0 = not in a profile"
```
in your yaml I think.  Let me know if that is not working

EllisK-L:
Oh sorry, I was more talking about the defaults. I know that you can specify PROFILE_NUMBER which will replace profile_index, same with the other calculated variables. What I was trying to get at is if people think it should automatically detect if it's an OG1 dataset and do the conversion without needing to specifying in the YAML.
One of the reasons I bring this up is because there is already a section that adds in OG1 specific variables automatically in seaexplorer.py:728.

jklymak:
The defaults for the ioos DAC are there for back-compatibility so as not to break old yaml workflows.    I'm pretty hesitant to make a bunch of defaults for OG1.0.  If anything I'd go the other way and suggest we should make the examples for non OG1.0 compliant with the "new" yaml files.

> One of the reasons I bring this up is because there is already a section that adds in OG1 specific variables automatically in seaexplorer.py:728.

Yeah, I'll admit that is a bit of a lazy hack.  I probably should try to get rid of that.

EllisK-L:
Okay, so you want the direction of pyglider to be as format agnostic as possible while of course not breaking existing workflows, that makes sense.
In that case I agree that there should be minimal defaults in the code for specific formats, and more heavily rely on the yamls. 
I was thinking that the specific cases mentioned in seaexplorer.py (*_GPS) be a configurable derived variable, but that might be a bit tricky since it's so platform dependent for how it's calculated.

PR #241, opened 2026-06-19
TST: yaml approach

jklymak:
Previously, we had netcdf files that were generated by the tests and compared to "golden" netcdf files.  Unfortunately, those change often, and often in ways that are hard to track.  

Here we use a different approach, whereby the "golden" state are yaml files generated from the test netcdf files, and saved as text files.  Data stats are kept by looking at mean, std, min/max, and the mean and std of the time derivative.  So if the data file changes substantially they should trigger an error.  

If the golden files change, then we run `python tests/_generate_expected_yaml.py` and it regenerates the yaml files, and the diffs will be in GitHub for posterity and easy inspection.  

If a file changes a lot, there will still be a need for the developer to manually inspect what changed...


jklymak:
I'm going to go ahead and merge these changes.  They should not affect processing data in any way, and I think make for a better CI and git environment.

PR #245, opened 2026-07-03
Slocum OG1.0 GPS Fixes Update

EllisK-L:
* Changed the Slocum processor for deriving `LATITUDE_GPS, LONGITUDE_GPS, TIME_GPS` to match how Seaexplorer does it (using the YAML).
* Added Linux-64 support in the project.toml

Something I also wanted to ask here, the yaml uses the processing method `gps_fixes_from_nav`. Since Slocum uses `m_gps_lat/lon` instead of `NavState` should the processing method name be more generic, like `gps_fixes`, or have slocum use a different name, maybe something like `gps_fixes_from_m_gps`,  or `gps_fixes_from_measured`

### DIFF
--- a/pyglider/slocum.py
+++ b/pyglider/slocum.py
@@ -1044,17 +1044,30 @@ def binary_to_timeseries(
 
     ds = utils.fill_metadata(ds, deployment['metadata'], device_data, varnames=varnames)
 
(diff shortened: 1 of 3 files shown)


jklymak (approved):
Thanks!

PR #248, opened 2026-07-07
gridding updates

smwoodman:
Addresses #226 

- fix bug to check for 'average_method' in `ds[k].attrs.keys()`, rather than in `ds[k].attrs.values()`. 
- calculate profile_direction using mode, and ensure is only has time dimension. 
- skip gridding variable if variable includes attribute 'average_method: none' in input timeseries. 
- distance_over_ground (or DISTANCE_OVER_GROUND) is never gridded. 
- if a variable does not have an average_method attribute, add 'average_method = arithmetic mean' in output netcdf. 
- update tests and docs


pyglider/ncprocess.py
```diff
@@ -360,31 +377,39 @@ def make_gridfiles(
...
 
     ds = ds.drop_vars('time_1970')
-    _log.info(f'Done times!')
+    _log.info('Done times!')
 
 
-    skip_vars = {'time', lat_varname, lon_varname, depth_varname, profile_index_varname}
+    skip_vars = {
+        'time', lat_varname, lon_varname, depth_varname, profile_index_varname, 
+        "distance_over_ground", "DISTANCE_OVER_GROUND",
```
smwoodman: I tried to do this in some way other than adding both 'distance_over_ground' and 'DISTANCE_OVER_GROUND' to `skip_vars` (eg, using `utils._resolve_role`), but couldn't figure out how

pyglider/ncprocess.py
```diff
@@ -323,19 +330,26 @@ def make_gridfiles(
...
+                ds[td].attrs["average_method"] = "mode"
+            else:
+                method = 'mean'
+                ds[td].attrs["average_method"] = "arithmetic mean"
+
             dat, xedges, binnumber = stats.binned_statistic(
                 ds[profile_index_varname].values[good],
                 ds[td].values[good],
-                statistic='mean',
+                statistic=method,
```
smwoodman, line 353: Used `method` to be consistent with the `for k in ds.keys():` block, even though the `for td, bin_stat in profile_lookup.items():` uses `bin_stat` as the varname instead. 

Update: I also switched from `bin_stat` to `method` in the `for td, bin_stat in profile_lookup.items():` block, to be consistent throughout the function.

pyglider/ncprocess.py
jklymak: I think if we look at the OG1 files we will see an entry for DISTANCE_OVER_GROUND

```
  DISTANCE_OVER_GROUND:
    processing_method:
      distance_over_ground:
        latitude: LATITUDE
        longitude: LONGITUDE
    long_name:     Distance over ground flown since mission start
    units:         km
```
I think this should gain a `processing_role: distance_over_ground` and then the "default" distance over ground should also get that processing role.

smwoodman: Makes sense. I added `processing_role: distance_over_ground` to the example slocum/seaexplorer OG1 yml files, and updated the gridding code to use `utils._resolve_role` for dog. 

> and then the "default" distance over ground should also get that processing role.

I think this means to add 'distance_over_ground' to the 'known_roles' block in [`_get_varnames`](https://github.com/c-proof/pyglider/blob/4fe4db0a5716087bb340155b7953d9d0ec5e3e55/pyglider/utils.py#L839), so I added it there (original shown below). But please redirect me as needed
```
def _get_varnames(deployment):
...
    known_roles = {
        'time', 'latitude', 'longitude', 'pressure', 'temperature',
        'conductivity', 'salinity', 'depth', 'profile_index',
        'profile_direction', 'oxygen_concentration',
    }
```

PR #251, opened 2026-08-12
FIX: make opening polars more robust

jklymak:
There seem to be issues reading SeaExplorer data on windows machines (https://github.com/c-proof/pyglider/discussions/250).  

This is an attempt to make more robust by opening as a binary stream first.  However, untested as I do not have access to a windows machine @Kourtney-Burger is there any chance you could try this patch?

### DIFF
--- a/pyglider/seaexplorer.py
+++ b/pyglider/seaexplorer.py
@@ -140,7 +140,8 @@ def raw_to_rawnc(
                     # Try to read the file with polars. If the file is corrupted (rare), file read will fail and file
                     # is appended to badfiles
                     try:
-                        out = pl.read_csv(f, separator=';')
+                        with open(f, 'rb') as fin:
+                            out = pl.read_csv(fin, separator=';')
                     except Exception as e:
                         _log.warning(f'Exception reading {f}: {e}')
                         _log.warning(f'Could not read {f}')
@@ -632,8 +633,13 @@ def raw_to_timeseries(
         )[0]
         + 1
     )
-    ds[lon_name].values = np.interp(ds.time, ds.time[good], ds[lon_name][good])
-    ds[lat_name].values = np.interp(ds.time, ds.time[good], ds[lat_name][good])
+    if len(good) >= 2:
+        ds[lon_name].values = np.interp(ds.time, ds.time[good], ds[lon_name][good])
+        ds[lat_name].values = np.interp(ds.time, ds.time[good], ds[lat_name][good])
+    else:
+        _log.warning(
+            'Not enough good lat/lon fixes to interpolate - using original values'
+        )
 
     # keep only timestamps with data from one of a set of variables
     if 'keep_variables' in ncvar:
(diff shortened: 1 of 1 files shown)


jklymak:
Can you include your processing script?  I'm not clear what you are trying to do with the `pld.sub` files if you already have `pld.raw` files.  Are you trying to do realtime processing or delayed?

jklymak:
Also you have supplied the subsetted payload files 10 and 11, but no glider/nav files.  The processing needs them to match.

Ideally you would supply exactly what script you ran and exactly the files needed to run it.

jklymak:
OK, so looking at `sea117.26.pld1.sub.10.gz` it is not a gzip file.  Did the glider make this?  It is instead a flat zip file:

```
file realtime_raw/sea117.26.pld1.sub.11.gz
realtime_raw/sea117.26.pld1.sub.11.gz: Zip archive data, at least v2.0 to extract, compression method=deflate
```

If you run unzip on it:

```
unzip realtime_raw/sea117.26.pld1.sub.11.gz 
Archive:  realtime_raw/sea117.26.pld1.sub.11.gz
  inflating: sea117.26.pld1.sub.11   
 extracting: sea117.26.11.sonics.spectros.xz  
```

So you can see it has two files embedded in it, one is an acoustics file, I assume, and the other is the rest of the payload data.  

1) whoever made this a zip file should have added a suffix `.zip` _not_ `.gz`.
2) you will have to unzip this manually for each of your files; this is beyond the purview of pyglider.  

If you sort out 1) it may be possible for pyglider to automatically detect that this has happened, but unless it is a known specification, which I doubt because of the misleading suffix, I;m not sure we should try and do anything

Kourtney-Burger:
Hi Jody, thanks for continuing to look into this. My [script is available here](https://github.com/NMFS-PAM-Glider/GliderRodeo/blob/SeaExplorer-Processing/_code/SEA117-M026_20260128/SEA117-M026_20260128-delayed.py) and the [updated test data is here](https://github.com/NMFS-PAM-Glider/GliderRodeo/tree/SeaExplorer-Processing/data/SEA117-M026_20260128/testFiles) (excluding sub 10 and 11 files). I'm doing delayed processing and it is working with the raw files so that's all good. 

I also ran the script on a remote linux workstation and saw the same messages in the processing log ([available here](https://github.com/NMFS-PAM-Glider/GliderRodeo/blob/SeaExplorer-Processing/data/SEA117-M026_20260128/SEA117-M026_20260128-processing.log)) that it could not read the pld.sub files. It makes sense that the incorrect suffix is causing problems. 

I don't think I can properly unzip the pld1.sub.##.gz files on my windows machine, when I try to extract it I only see the sea117.26.pld1.sub.11 file and it is unreadable. I can unzip it with my linux workstation and see the two files like you mentioned. I didn't know linux worked differently with this so thanks for including that explanation! 

I received those files from Alseamar, I was not involved in the data offload process so I am not sure if this compression suffix was from the realtime data offloaded from glimpse or from somewhere else. Like you mentioned, because I have the pld.raw I am not too worried about this issue now. 

My next steps will be working on the level-1 timeseries processing. I can follow up with more processing questions on the [original discussion](https://github.com/c-proof/pyglider/discussions/250). 

thanks!

